PR #28, opened 2021-02-13
v0.2.4

sonichi:
sample_weight; dependency; notebook


flaml/automl.py
```diff
@@ -1151,7 +1153,11 @@ def _search(self):
                 stacker = Stacker(estimators, best_m,
                     n_jobs=self._state.n_jobs,
                     passthrough=True)
-                stacker.fit(self._X_train_all, self._y_train_all)
+                if self._sample_weight_full is not None:
```
qingyun-wu, line 1156: Why line 1156-1158 is needed? Or in other words why self._sample_weight_full is needed? why can't just use self._state.fit_kwargs.get('sample_weight')?

sonichi: `self._state.fit_kwargs['sample_weight']` was revised in `_prepare_data()`: https://github.com/microsoft/FLAML/blob/d18d292081dbf34ffb2b1e3948acaf37569ebedd/flaml/automl.py#L477

PR #30, opened 2021-02-18
v0.2.5

sonichi:
None check in set_search_properties


flaml/searcher/flow2.py
```diff
@@ -237,7 +239,8 @@ def complete_config(self, partial_config: Dict,
         ''' generate a complete config from the partial config input
         add minimal resource to config if available
         '''
-        if self._reset_times: # not the first time, use random gaussian
+        if self._reset_times and partial_config==self.init_config:
```
qingyun-wu, line 242: why the second condition is needed? can you add comments?

flaml/searcher/blendsearch.py
```diff
@@ -314,7 +313,6 @@ def suggest(self, trial_id: str) -> Optional[Dict]:
                 self._result[config_signature] = {}
             else: return None # running but no result yet
             self._init_used = True
-            self._trial_proposed_by[trial_id] = 0
```
qingyun-wu, line 317: why removing this?

flaml/searcher/flow2.py
sonichi: added

qingyun-wu (approved):

PR #32, opened 2021-02-23
v0.2.6

sonichi:
xgboost notebook


flaml/searcher/search_thread.py
```diff
@@ -70,7 +71,7 @@ def _update_speed(self):
         # calculate speed; use 0 for invalid speed temporarily
         if self.obj_best2 > self.obj_best1: 
             self.speed = (self.obj_best2 - self.obj_best1) / (
-                self.cost_total - self.cost_best2)
+                self.cost_total - self.cost_best2 + self.eps)
```
qingyun-wu, line 74: why adding self.eps? (presumably self.cost_total should not equal to self.cost_best2)

sonichi: test_nested fails in test_tune.py with DivideByZeroException

qingyun-wu: why does it fail?

sonichi: self.cost_total = self.cost_best2

qingyun-wu: So we have self.obj_best2 > self.obj_best1 and  self.cost_total = self.cost_best2 at the same time?

sonichi: yes

qingyun-wu (approved):

qingyun-wu (approved):

PR #35, opened 2021-03-03
V0.2.7

sonichi:
fix bug in BlendSearch and FLOW^2.


flaml/searcher/blendsearch.py
```diff
@@ -279,28 +279,29 @@ def suggest(self, trial_id: str) -> Optional[Dict]:
                 # logger.debug(f"random config {config}")
                 skip = self._should_skip(choice, trial_id, config)
```
qingyun-wu, line 280: why don't we break until we find one that we don't need to skip? It may get stuck in the for loop?

flaml/searcher/blendsearch.py
```diff
@@ -279,28 +279,29 @@ def suggest(self, trial_id: str) -> Optional[Dict]:
...
                 if skip: return None
-            # if not choice: logger.info(config)
-            if choice or backup == choice or self._valid(config): 
+            # if not choice: print(config)
+            if choice or self._valid(config): 
                 # LS or valid or no backup choice
                 self._trial_proposed_by[trial_id] = choice
             else: # invalid config proposed by GS
                 if not self._use_rs:
```
qingyun-wu, line 287: why the condition is 'not self._use_rs'?  Shouldn't it be 'self._use_rs'?

flaml/searcher/blendsearch.py
```diff
@@ -279,28 +279,29 @@ def suggest(self, trial_id: str) -> Optional[Dict]:
...
                     self._search_thread_pool[choice].on_trial_complete(
                         trial_id, {}, error=True) # tell GS there is an error
                 self._use_rs = False
+                if choice == backup: return None
```
qingyun-wu, line 291: why return None?

flaml/searcher/blendsearch.py
```diff
@@ -315,6 +316,7 @@ def suggest(self, trial_id: str) -> Optional[Dict]:
                 self._result[config_signature] = {}
             else: return None # running but no result yet
             self._init_used = True
+            self._trial_proposed_by[trial_id] = 0
```
qingyun-wu, line 319: consider init to be proposed by GS?

flaml/searcher/flow2.py
```diff
@@ -164,6 +164,7 @@ def _init_search(self):
                 if str(sampler) != 'Normal':
                     self._bounded_keys.append(key)
         self._space_keys = list(self.space.keys())
+        print(self._ordered_choice_hp)
```
qingyun-wu, line 167: remove print?

flaml/searcher/flow2.py
```diff
@@ -306,10 +307,10 @@ def normalize(self, config) -> Dict:
                         # normalize categorical
                         if key in self._ordered_cat_hp:
                             l, d = self._ordered_cat_hp[key]
-                            config_norm[key] = d[value]/len(l)
+                            config_norm[key] = (d[value]+0.5)/len(l) # center
```
qingyun-wu, line 310: why?

flaml/searcher/search_thread.py
```diff
@@ -55,7 +57,7 @@ def suggest(self, trial_id: str) -> Optional[Dict]:
 
     def update_priority(self, eci: Optional[float] = 0):
         # optimistic projection
-        self.priority = eci * self.speed - self.obj_best1
+        self.priority = (eci * self.speed - self.obj_best1)
```
qingyun-wu, line 60: why brackets are needed?

flaml/searcher/search_thread.py
```diff
@@ -82,7 +84,11 @@ def on_trial_complete(self, trial_id: str, result: Optional[Dict] = None,
         if not hasattr(self._search_alg, '_ot_trials') or (not error and
             trial_id in self._search_alg._ot_trials):
             # optuna doesn't handle error
-            self._search_alg.on_trial_complete(trial_id, result, error)
+            if self._is_ls:
+                self._search_alg.on_trial_complete(trial_id, result, error)
+            elif self._init_config: self._init_config = False
+            else: # init config is not proposed by self._search_alg
+                self._search_alg.on_trial_complete(trial_id, result, error)
```
qingyun-wu, line 91: why write if else in this way?

flaml/searcher/flow2.py
```diff
@@ -409,6 +410,7 @@ def set_search_properties(self,
             self._metric = metric
         if mode:
             assert mode in ["min", "max"], "`mode` must be 'min' or 'max'."
+            self._mode = mode
```
qingyun-wu, line 413: self.mode? or need to change self.mode in line 287 to self._mode

qingyun-wu (changes requested):

flaml/searcher/search_thread.py
sonichi: updated

flaml/searcher/search_thread.py
sonichi: removed

flaml/searcher/flow2.py
sonichi: because we use floor() to convert from float to int.

flaml/searcher/flow2.py
sonichi: done

flaml/searcher/blendsearch.py
sonichi: to update the search thread states.

flaml/searcher/blendsearch.py
sonichi: changed to use CFO's init

flaml/searcher/blendsearch.py
sonichi: yes, to avoid getting stuck.

qingyun-wu (approved):

PR #37, opened 2021-03-09
pickle the AutoML object

sonichi:
@ekzhu I set self.file=None after close() is called for making the AutoML object picklable. Could you review it?


ekzhu (commented):
Looks good to me. Only note is that there should be a version number for pickled object because the pickled object will be outdated as the class evolves overtime. So when you unpickle an serialized object you need to compare its version with the current running version of AutoML. See example: https://stackoverflow.com/questions/2361140/pickled-object-versioning

sonichi:
> Looks good to me. Only note is that there should be a version number for pickled object because the pickled object will be outdated as the class evolves overtime. So when you unpickle an serialized object you need to compare its version with the current running version of AutoML. See example: https://stackoverflow.com/questions/2361140/pickled-object-versioning

@ekzhu thanks. Could you check the updated version?

flaml/automl.py
```diff
@@ -1208,9 +1226,11 @@ def _select_estimator(self, estimator_list):
                 gap = search_state.best_loss - self._state.best_loss
                 if gap > 0 and not self._ensemble:
                     delta_loss = (search_state.best_loss_old - 
-                                 search_state.best_loss)
+                                 search_state.best_loss) or \
+                                     search_state.best_loss
                     delta_time = (search_state.total_time_used - 
-                                 search_state.time_best_found_old)
+                                 search_state.time_best_found_old) or (
+                                     search_state.total_time_used + 1e-10)
```
qingyun-wu, line 1233: is this correct? in this case the 'real' delta_time, i.e., (search_state.total_time_used - search_state.time_best_found_old) is 0, do we really want to set delta_time to be search_state.total_time_used + 1e-10? shouldn't set it as 1e-10?

flaml/searcher/blendsearch.py
```diff
@@ -143,20 +144,31 @@ def _init_search(self):
         self._deadline = np.inf
 
     def save(self, checkpoint_path: str):
-        save_object = (self._metric_target, self._search_thread_pool,
-            self._thread_count, self._init_used, self._trial_proposed_by,
-            self._ls_bound_min, self._ls_bound_max, self._result,
-            self._deadline)
+        save_object = self
```
qingyun-wu, line 147: Does it mean to save the whole BlendSearch instance?

flaml/searcher/blendsearch.py
```diff
@@ -526,3 +538,87 @@ def _create_condition(self, result: Dict) -> bool:
         return len(self._search_thread_pool) < 2
 
 
+def create_next(client):
```
qingyun-wu, line 541: What is the type of the input client?

flaml/searcher/blendsearch.py
sonichi: yes, to ensure all the states are saved.

flaml/searcher/blendsearch.py
sonichi: not defined in this repo

flaml/automl.py
sonichi: Thanks. Fixed.

flaml/automl.py
```diff
@@ -1208,9 +1226,11 @@ def _select_estimator(self, estimator_list):
                 gap = search_state.best_loss - self._state.best_loss
                 if gap > 0 and not self._ensemble:
                     delta_loss = (search_state.best_loss_old - 
-                                 search_state.best_loss)
+                                 search_state.best_loss) or \
+                                     search_state.best_loss
```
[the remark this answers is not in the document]
sonichi: As defined in the paper.

qingyun-wu (approved):

Issue #38, opened 2021-03-10
Verbose argument in model.fit()


rroy09:
Hi,

While training the learner, a console output is generated, which can take up huge space in the notebook if the time_budget is made large. If I wish to suppress the console output while training my learner, how do I do that? In keras, sklearn, etc., setting verbose = 0 suppresses the console output.

Thanks!

sonichi:
Thanks for the question. The easiest way to do it in your notebook now is to insert these two lines before `automl.fit()`:

`from flaml import logger`
`logger.setLevel(logging.WARNING)`

for example:
`from flaml import logger`
`logger.setLevel(logging.WARNING)`
`automl.fit(X_train = X_train, y_train = y_train, **settings)`

I'd also encourage a pull request to add an argument `verbose` in `automl.fit()`. Feel free to do it if you'd like to contribute.

rroy09:
Thanks for the response. However, I get an error 'logging is not defined'.

sonichi:
@rroy09 Please also add `import logging`.

rroy09:
Thanks a lot for the response.

PR #40, opened 2021-03-12
Update automl.py to add verbose argument to fit

ImanHosseini:
as per: https://github.com/microsoft/FLAML/issues/38
Adding _verbose_ to fit: the arg is passed down to _search (instead of calling _search with verbose=0) also sets logging level to WARNING and if verbose = 0 and sets it back before returning. 

[GitHub is having issues today! It messed up my previous PR]


flaml/automl.py
```diff
@@ -823,6 +824,8 @@ def custom_metric(X_test, y_test, estimator, labels,
             y_val: None | a numpy array or a pandas series of validation labels
             sample_weight_val: None | a numpy array of the sample weight of
                 validation data
+            verbose : int, default=0 | Controls the verbosity, higher means more
```
sonichi, line 827: default=1?

flaml/automl.py
```diff
@@ -906,6 +913,8 @@ def custom_metric(X_test, y_test, estimator, labels,
             self._state.n_jobs = n_jobs
             self._search()
             logger.info("fit succeeded")
+        if verbose:
```
sonichi, line 916: what if verbose==0?

flaml/automl.py
```diff
@@ -1018,7 +1027,7 @@ def _search(self):
                 init_config=None, 
                 search_alg=search_state.search_alg,
                 time_budget_s=budget_left,
-                verbose=0, local_dir='logs/tune_results',
+                verbose=self.verbose, local_dir='logs/tune_results',
```
sonichi, line 1030: Could you test run `notebook/flaml_automl` to see if the output is desirable when verbose=1? If you are not sure please copy a snippet of the output from it and paste here.

sonichi (changes requested):

flaml/automl.py
sonichi: remove the extra space before colon

flaml/automl.py
ImanHosseini: I ran the notebook, if verbose is 0, what I get is:
"""
C:\Python\Python3.7\WPy64-3771\python-3.7.7.amd64\lib\site-packages\xgboost\sklearn.py:888: UserWarning: The use of label encoder in XGBClassifier is deprecated and will be removed in a future release. To remove this warning, do the following: 1) Pass option use_label_encoder=False when constructing XGBClassifier object; and 2) Encode your labels (y) as integers starting with 0, i.e. 0, 1, 2, ..., [num_class - 1]. 
"""
(multiple of these messages, because it is called on every pass, but no 'INFO' statements)
When I run with verbose = 1 I get the warnings AND I get these:
"""
[2021-03-13 14:10:53] INFO (flaml.tune.tune/MainThread) trial 1 config: {'n_estimators': 4.0, 'max_leaves': 8.0, 'min_child_weight': 1.4505165368937825, 'learning_rate': 0.12354235042058405, 'subsample': 0.9504860919941635, 'log_max_bin': 7.0, 'colsample_bytree': 0.9658058260875579, 'reg_alpha': 2.955773383057842e-10, 'reg_lambda': 0.23033924766285224, 'FLAML_sample_size': 10000}
"""
(and all other 'INFO' statements, basically verbose = 1 is as it was before this change, verbose = 0 now adds the possibility to silence those 'INFO' statements)

flaml/automl.py
ImanHosseini: So before adding the verbose arguments, 'INFO' statements were being printed (as if verbose is >0). but at the same time, when calling _search(...) , 0 was being passed to verbose. Now we either pick verbose default value to be 1 (so that we get those 'INFO' prints) or if we set verbose=0, then due to some previous test that checks debug buffer, we would fail that test, and then we would have to change that test.

flaml/automl.py
ImanHosseini: my bad, this should be "if verbose==0" (so that if we have previously changed logger level, we revert it to what it was before.

flaml/automl.py
sonichi: My desired default behavior is: output the "INFO" in flaml; don't output the "INFO" in flaml.tune. What about the following?
default: `verbose=1`, which output "INFO" in flaml but don't output "INFO" in flaml.tune.
`verbose=0`: output no "INFO" at all.
`verbose=2`: output "INFO" in both flaml and flaml.tune.

flaml/automl.py
sonichi: Please my last comment in the bottom.

flaml/automl.py
ImanHosseini: Nice! Because tune handles verbose flag already: https://github.com/microsoft/FLAML/blob/main/flaml/tune/tune.py#L227
The only thing is, tune would be at 'DEBUG' for verbose=3 and 'INFO' for verbose in {2, 1} so we only need to pass verbose-1 to tune.run and we're good.

flaml/automl.py
```diff
@@ -823,6 +824,8 @@ def custom_metric(X_test, y_test, estimator, labels,
             y_val: None | a numpy array or a pandas series of validation labels
             sample_weight_val: None | a numpy array of the sample weight of
                 validation data
+            verbose: int, default=0 | Controls the verbosity, higher means more
```
sonichi, line 827: default=1

sonichi (changes requested):
looks good except the default value

flaml/automl.py
ImanHosseini: ouch, sorry!

sonichi (approved):

qingyun-wu (approved):

PR #42, opened 2021-03-17
Normalize config in _valid

Yard1:
`_valid` method in BlendSearch does not normalize the config it recieves, leading to denormalized values being compared to normalized min and max values and thus almost always failing the check. This PR fixes that.


sonichi (approved):
Thanks for the fix.

Issue #36, opened 2021-03-08
Write example test for using nni tuning interface


sonichi:
add test/nni to illustrate how to use flaml in the nni framework.

ImanHosseini:
I tried to use the _BlendSearchTuner_ in NNI, for the mnist case, I tried this config.yaml:
```yaml
authorName: default
experimentName: example_mnist
trialConcurrency: 1
maxExecDuration: 1h
maxTrialNum: 10
trainingServicePlatform: local
# The path to Search Space
searchSpacePath: search_space.json
useAnnotation: false
tuner:
  codeDir: /path/to/flaml/searcher/
  classFileName: blendsearch.py
  className: BlendSearchTuner
# The path and the running command of trial
trial:
  command: python3 mnist.py
  codeDir: .
  gpuNum: 0
```
But it crashes with this error:
```
[2021-03-13 15:04:12] ERROR (nni.main/MainThread) attempted relative import with no known parent package
Traceback (most recent call last):
  File "/home/iman/.local/lib/python3.8/site-packages/nni/__main__.py", line 108, in <module>
    main()
  File "/home/iman/.local/lib/python3.8/site-packages/nni/__main__.py", line 43, in main
    tuner = _create_tuner(exp_params)
  File "/home/iman/.local/lib/python3.8/site-packages/nni/__main__.py", line 87, in _create_tuner
    tuner = create_customized_class_instance(exp_params.get('tuner'))
  File "/home/iman/.local/lib/python3.8/site-packages/nni/tools/package_utils/__init__.py", line 200, in create_customized_class_instance
    class_module = importlib.import_module(module_name)
  File "/usr/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 783, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/mnt/c/Users/salva/Documents/GitHub/FLAML/test/../flaml/searcher/blendsearch.py", line 17, in <module>
    from .search_thread import SearchThread
ImportError: attempted relative import with no known parent package
[...]

PR #44, opened 2021-03-18
nni example + bugfix in blendsearch

ImanHosseini:
added an NNI example and fixed a bug in blendsearch.
re: https://github.com/microsoft/FLAML/issues/36


sonichi (approved):
please have a new line at the end of each file and then it's good to merge.

sonichi (approved):

tanmay2798:
Hi, I had a question, there are cases when None is returned as configuration from suggest function of Blendsearch. So in that case, nni.get_next_parameter() receives None configuration. How can we handle that case?

PR #49, opened 2021-03-20
nni example

sonichi:


qingyun-wu (approved):

tanmay2798:
Hi, I had a question, there are cases when None is returned as configuration from suggest function of Blendsearch. So in that case, nni.get_next_parameter() receives None configuration. How can we handle that case?

sonichi:
I believe None is a valid output expected by nni.get_nex_parameter(). Please correct me if I'm wrong.

tanmay2798:
I think that by default nni.get_nex_parameter() would never return None when another tuner is used like TPE or Anneal or so on. But when Blendsearch is used, it returns None as a configuration also.
So basically if nni.get_nex_parameter() return None, then there would be no results corresponding to that trial. But this is never the case in other tuners as they are designed to stop when no more trials are available.

tanmay2798:
Can you provid me your email id, so maybe we can discuss this in depth.

sonichi:
> I think that by default nni.get_nex_parameter() would never return None when another tuner is used like TPE or Anneal or so on. But when Blendsearch is used, it returns None as a configuration also. So basically if nni.get_nex_parameter() return None, then there would be no results corresponding to that trial. But this is never the case in other tuners as they are designed to stop when no more trials are available.

I see. Then it is an issue. Would you like to chat more on discord? https://discord.gg/Cppx2vSPVP

PR #67, opened 2021-04-19
(data.py) fit_transform method able to transform datatime columns to float values.

gianpDomiziani:
sklearn\utils\validation.py needs float/int values. 
fit_transform method (in data.py module) now check if some datetime columns is present, if yes , they are transformed to float types.

### DIFF
--- a/flaml/data.py
+++ b/flaml/data.py
@@ -196,8 +196,8 @@ class DataTransformer:
             drop = False
             for column in X.columns:
                 # sklearn\utils\validation.py needs int/float values
-                X.loc[:, column] = X[column].map(datetime.toordinal) if X[column].dtype.name in (
-                    'datetime64[ns]') else X[column]
+                if X[column].dtype.name == 'datetime64[ns]':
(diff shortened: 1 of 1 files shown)


flaml/data.py
```diff
@@ -193,6 +195,9 @@ def fit_transform(self, X, y, task):
             cat_columns, num_columns = [], []
             drop = False
             for column in X.columns:
+                # sklearn\utils\validation.py needs int/float values
+                X.loc[:, column] = X[column].map(datetime.toordinal) if X[column].dtype.name in (
```
sonichi, line 199: I suggest a small change:
```python
if X[column].dtype.name == 'datetime64[ns]':
    X[column] = X[column].map(datetime.toordinal)
```
I assumed you meant a tuple when using `('datetime64[ns]')`.

gianpDomiziani: Sure, it is better!

sonichi (approved):
You can increase the version number to 0.3.2.

qingyun-wu (approved):

Issue #66, opened 2021-04-19
datetime64[ns] dtype in dataframe


gianpDomiziani:
When, DataTranformer's fit_transform method is called,  If some columns have a datetime format, an error is raised by the sklearn\utils\validation.py method. 

I fixed it, turning any datetime columns to datetime.toordinal type

sonichi:
Thanks @gianpDomiziani. This a reasonable approach before we add preprocessing specific to datetime type. Would you like to create a PR with your fix?

PR #68, opened 2021-04-20
 (data.py) fit_transform method able to transform datatime columns to float values. 

gianpDomiziani:
sklearn\utils\validation.py needs float/int values.
fit_transform method (in data.py module) now check if some datetime columns is present, if yes , they are transformed to float types.

(sorry, i accidentaly closed the old pull request)


sonichi (approved):

sonichi:
close #66

PR #77, opened 2021-04-23
directly check if X_val needs datatime preprocess.

gianpDomiziani:
the transform method is called again before the prediction call, so a directly check is needed.

### DIFF
--- a/flaml/data.py
+++ b/flaml/data.py
@@ -214,6 +214,7 @@ class DataTransformer:
                         X[column] = X[column].fillna('__NAN__')
                         cat_columns.append(column)
                 else:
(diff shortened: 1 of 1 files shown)


sonichi (commented):
Could you provide one example why this check is needed?

gianpDomiziani:
**That pull request, is not more required**.

The problem was that if you use automl.predict(X_val) after the fitting  this error arise: TypeError: descriptor 'toordinal' for 'datetime.date' objects doesn't apply to a 'int' object. This beacuse, in the transformer method, it is used directly the passed X, without taking a copy of it. I resolve by copy the passed X.

PR #78, opened 2021-04-23
X.copy() in the process method

gianpDomiziani:
X.copy() is needed in the transformer method, otherwise the passed X_val is changed at the fittining phase, therefore if the predict is called, the X_val has not more datetime columns, due to the before transform.


flaml/data.py
```diff
@@ -249,9 +250,10 @@ def fit_transform(self, X, y, task):
         return X, y
 
     def transform(self, X):
+        X = X.copy()
         if isinstance(X, pd.DataFrame):
-            cat_columns, num_columns, datetime_columns = \
-                self._cat_columns, self._num_columns, self._datetime_columns
+            cat_columns, num_columns, datetime_columns = self._cat_columns, \
+                                                         self._num_columns, self._datetime_columns
             if datetime_columns:
```
sonichi, line 257: could you move line 260 before 257 and remove all the other changes?

sonichi (changes requested):

sonichi (changes requested):
could you change the version number to 0.3.4 in __version__.py?

sonichi (approved):

PR #88, opened 2021-05-15
constraints

sonichi:
generalize the mem constraints to generic constraints


test/tune/test_tune.py
```diff
@@ -163,21 +166,28 @@ def test_nested():
     }
 
     def simple_func(config):
-        tune.report(metric=(config["cost_related"]["a"] - 4)**2
-                    * (config["b"] - 0.7)**2)
+        metric = (config["cost_related"]["a"] - 4)**2 \
```
qingyun-wu, line 169: change 'metric' to 'obj'

flaml/version.py
```diff
@@ -1 +1 @@
-__version__ = "0.3.6"
```
qingyun-wu, line 1: do not change version

flaml/searcher/blendsearch.py
```diff
@@ -39,9 +39,11 @@ def __init__(self,
                  min_resource: Optional[float] = None,
                  max_resource: Optional[float] = None,
                  reduction_factor: Optional[float] = None,
-                 resources_per_trial: Optional[dict] = None,
                  global_search_alg: Optional[Searcher] = None,
-                 mem_size: Callable[[dict], float] = None,
+                 config_constraints: Optional[
+                     List[Tuple[Callable[[dict], float], str, float]]] = None,
```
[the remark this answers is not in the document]
sonichi: list of arguments

flaml/version.py
sonichi: changing version number doesn't requires publishing to pypi.

test/tune/test_tune.py
sonichi: done

flaml/searcher/blendsearch.py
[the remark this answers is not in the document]
aminsaied: This is hard to parse: List[Tuple[Callable[[dict], float], str, float]]]

What about introducing a lightweight dataclass to manage these? e.g.

@dataclass
class Constraint:
  metric_function: Callable[Dict, float]
  inequality: str
  value: float

and replace type with Optional[List[Constraint]]. This would also make the code more readable below.

qingyun-wu (approved):

PR #99, opened 2021-06-03
Multiclass

### LINKED ISSUE #97: FLAML: New feature
My team is working on a multiclass classification model for a predicting the workload type(POC, Prod, Dev, Test) of Azure services like SQLDW, Synapse and SQLDB. We replaced the Gridsearch/XGBoost with FLAML XGBoost for better performance. Since it is multiclass classification, we implemented more metrics like normalized confusion matrix, Precision-Recall curve and Roc-curve using OneVsRestClassifier for binarizing the labels for our final model so that we can measure the performance for prediction of each individual workload type, in additional to accuracy, precision and recall of overall model. This seems like a common requirement that other FLAML users might have and it will be valuable to add these features for multiclass classification models.

The link to access the jupyter notebook for multiclass classification is 
https://microsoft.sharepoint.com/:u:/t/AzureDataUXBA-DataEngineeringandAnalysis/ETY_DWyvPXBEl2S-R5C6rVUBFa0fvbnE9V7KSzAC3H8uMQ?e=hgxKmb

It has the implementation of above metrics in the last section of the file(5. Metrics).

sonichi:
#97 produce normalized confusion matrix and roc curves or precision-recall curves for each class


flaml/searcher/flow2.py
```diff
@@ -124,15 +124,14 @@ def _init_search(self):
...
-                #     if str(sampler_inner) == 'Uniform':
-                #         self._step_lb = min(
-                #             self._step_lb, sampler.q/(domain.upper-domain.lower))
-                # elif isinstance(domain, sample.Integer) and str(
-                #     sampler) == 'Uniform':
-                #     self._step_lb = min(
-                #         self._step_lb, 1.0/(domain.upper-domain.lower))
+                if isinstance(sampler, sample.Quantized):
+                    sampler_inner = sampler.get_sampler()
+                    if str(sampler_inner) == 'Uniform':
```
qingyun-wu, line 129: why only handling 'Uniform'? What about 'LogUniform'? 

and why in step_lower_bound(), only 'LogUniform' is handled?

qingyun-wu (approved):

PR #102, opened 2021-06-05
move import position

sonichi:
to avoid import error of vw


flaml/onlineml/trial.py
```diff
@@ -306,7 +304,12 @@ def __init__(self,
                          custom_trial_name, self.trial_id)
         self.model = None   # model is None until the config is scheduled to run
         self.result = None
-        self.trainable_class = self.MODEL_CLASS
+        try:
+            from vowpalwabbit import pyvw
+        except ImportError:
+            raise ImportError(
+                'To use AutoVW, please run pip install flaml[vw] to install vowpalwabbit')
+        self.trainable_class = pyvw.vw
```
qingyun-wu, line 312: is this right?

sonichi: previously it was
MODEL_CLASS = pyvw.vw
self.trainable_class = self.MODEL_CLASS 
what problem do you think there is?

qingyun-wu (approved):

PR #107, opened 2021-06-13
groups

### LINKED ISSUE #56: Option for groupKFold for regression problems
Hi,

I'm trying to tune lightgbm for a regression problem and need to use groupKFold for cross-validation. 
By default, automl.fit() takes repeatedkfold as split_type. I looked up at the [documentation](https://microsoft.github.io/FLAML/), but couldn't find details regarding that. Also, how to pass the groups arguments to it. 

Thanks in advance.

sonichi:
Add an argument 'groups' to `AutoML.fit()`. Use GroupKFold when this argument is passed and the eval method is cv.

close #56


dtquandt:
Wow, thank you for this. I'm going to test it and report back if I find any issues.

dtquandt:
Seems to be working for my binary classification problem with both CV and holdout methods. Metrics progression seem appropriate for grouping as if it wasn't doing it it would be overfitting pretty much instantly. Will test with regression soon.

dtquandt:
It doesn't seem to be behaving any differently when I pass in groups (which are a pandas series of string names, but I've also tried a numpy array of strings, and a numpy array of integer category codes with no success.) It's hard to tell exactly what's happening under the hood as I've tried extracting the GroupKFold code from the FLAML implementation and it seems to do what it's supposed to, but I'm getting results that I know to be impossible unless there's data leakage to overfit on. Can you think of any reason why this would happen or steps to understand what's going on?

For context, each group is a match in a sport with many rows of data for that match, and without correctly separating each match into its own group it'll very easily overfit on identifying information. With the split, however, it shouldn't be able to do that. It works fine with sklearn's GridSearchCV/RandomizedSearchCV, so I'm not sure why it wouldn't here.

dtquandt:
From what I can tell it seems to be working as expected now!

qingyun-wu (approved):

PR #122, opened 2021-06-26
fixing "discount running thread "

liususan091219:


setup.py
```diff
@@ -55,7 +55,8 @@
             "wandb==0.10.26",
             "torch==1.8.1",
             "datasets==1.4.1",
-            "azure-storage-blob"
+            "azure-storage-blob",
+            "tensorflow"
```
sonichi, line 59: why is tensorflow required?

sonichi (approved):
Let's figure out whether tf is indeed necessary in the next PR.

PR #123, opened 2021-06-26
unordered categorical

sonichi:
rounding bug


flaml/searcher/flow2.py
```diff
@@ -190,7 +190,8 @@ def _init_search(self):
         # record intermediate trial cost
         self._trial_cost = {}
         self._same = False  # whether the proposedd config is the same as best_config
-        self._init_phrase = True  # initial phase to increase initial stepsize
+        self._init_phase = True  # initial phase to increase initial stepsize
+        self._trunc = 0
```
qingyun-wu, line 194: Set it as None?
```suggestion
        self._trunc = None
```

flaml/searcher/flow2.py
```diff
@@ -494,8 +501,10 @@ def on_trial_result(self, trial_id: str, result: Dict):
             # record the cost in case it is pruned and cost info is lost
             self._trial_cost[trial_id] = cost
 
-    def rand_vector_unit_sphere(self, dim) -> np.ndarray:
+    def rand_vector_unit_sphere(self, dim, trunc=1) -> np.ndarray:
         vec = self._random.normal(0, 1, dim)
+        if 0 < trunc < dim:
```
qingyun-wu, line 506:
```suggestion
        if trunc:
```

flaml/searcher/flow2.py
```diff
@@ -494,8 +501,10 @@ def on_trial_result(self, trial_id: str, result: Dict):
...
+            vec[np.abs(vec).argsort()[:dim - trunc]] = 0
```
qingyun-wu, line 507:
```suggestion
            vec[np.abs(vec).argsort()[:max(dim - trunc, 0)]] = 0
```

flaml/searcher/flow2.py
```diff
@@ -555,25 +565,46 @@ def suggest(self, trial_id: str) -> Optional[Dict]:
...
+            # random
+            for i, key in enumerate(self._tunable_keys):
+                if self._direction_tried[i] != 0:
+                    for _, generated in generate_variants({'config': {
+                        key: self.space[key]
+                    }}):
+                        if generated['config'][key] != best_config[key]:
+                            config[key] = generated['config'][key]
+                            return unflatten_dict(config)
+                        break
```
qingyun-wu, line 599: why adding break here?

flaml/searcher/flow2.py
```diff
@@ -555,25 +565,46 @@ def suggest(self, trial_id: str) -> Optional[Dict]:
...
+        else:
+            # check if config == best_config
```
qingyun-wu, line 601: What is this for?

flaml/searcher/flow2.py
```diff
@@ -494,8 +501,10 @@ def on_trial_result(self, trial_id: str, result: Dict):
             # record the cost in case it is pruned and cost info is lost
             self._trial_cost[trial_id] = cost
 
-    def rand_vector_unit_sphere(self, dim) -> np.ndarray:
+    def rand_vector_unit_sphere(self, dim, trunc=1) -> np.ndarray:
```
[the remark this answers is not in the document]
sonichi: I'll use 0

flaml/searcher/flow2.py
sonichi: I added comments about self._trunc

flaml/searcher/flow2.py
sonichi: same as before

flaml/searcher/flow2.py
sonichi: If the rounded config is equal to the best config, then the incumbent would not be moved. By moving the incumbent, we increase the chance to escape that local point.

sonichi:
close #125 #126

qingyun-wu (approved):

PR #143, opened 2021-07-30
None type object cannot call close

sek788432:
if the log_file_name argument is wrong, self.file is None.
And it will raise none type object cannot call close()


flaml/training_log.py
```diff
@@ -159,6 +161,8 @@ def training_log_writer(filename: str):
         w = TrainingLogWriter(filename)
         w.open()
         yield w
+    except FileNotFoundError:
```
ekzhu, line 164: We can remove this except block as the error message below doesn't provide more detailed information than the builtin error message.

flaml/training_log.py
```diff
@@ -169,5 +173,7 @@ def training_log_reader(filename: str):
         r = TrainingLogReader(filename)
         r.open()
         yield r
+    except FileNotFoundError:
```
ekzhu, line 176: Same here.

ekzhu (commented):
Thanks for the contribution! The error messages for file not exists is not much different from the builtin message. Perhaps we can remove the except blocks.

flaml/training_log.py
sek788432: I got it. It seems reasonable. Do I need to update another commit to modify it?

ekzhu: Yes. Please update with another commit. Thanks!

sek788432: I updated. Thanks!

ekzhu (approved):
Thanks!

Issue #142, opened 2021-07-29
Error when the argument `training_function` passed to `tune.run` returns a value


eduardobull:
Thank you for this great library!

I'm getting the following error when the argument `training_function` passed to `tune.run` returns a value:

```python
.venv\lib\site-packages\flaml\tune\tune.py in run(training_function, config, points_to_evaluate, low_cost_partial_config, cat_hp_cost, metric, mode, time_budget_s, prune_attr, min_resource, max_resource, reduction_factor, report_intermediate_result, search_alg, verbose, local_dir, num_samples, resources_per_trial, config_constraints, metric_constraints, use_ray)
    330             if result is not None:
    331                 if isinstance(result, dict):
--> 332                     tune.report(**result)
    333                 else:
    334                     tune.report(_metric=result)

UnboundLocalError: local variable 'tune' referenced before assignment
```

Code to reproduce the problem:
```python
# require: pip install flaml[blendsearch]
from flaml import tune
import time

def evaluate_config(config):
    metric = (round(config['x'])-85000)**2 - config['x']/config['y']
    return {"metric": metric}  # <----- HERE!

analysis = tune.run(
    evaluate_config,
    config={
        'x': tune.lograndint(lower=1, upper=100000),
        'y': tune.randint(lower=1, upper=100000)
    }
)
```

Maybe we can fix this by changing

https://github.com/microsoft/FLAML/blob/15fd8adac4738074c73d651691df6bc562fffa7d/flaml/tune/tune.py#L332-L334

to

```python
    report(**result) 
else: 
    report(_metric=result) 
```

sonichi:
Great suggestion. That should work. If you'd like to make a PR, please go ahead. Otherwise we'll fix it in the next version.

PR #145, opened 2021-07-30
fix UnboundLocalError in tune.run (#142)

eduardobull:
Fix UnboundLocalError exception in `tune.run` when `training_function` returns a value.

### DIFF
--- a/test/tune/test_tune.py
+++ b/test/tune/test_tune.py
@@ -217,6 +217,37 @@ def test_nested():
     logger.info(f"BlendSearch best result: {best_trial.last_result}")
 
 
+def test_run_training_function_return_value():
+    from flaml import tune
+
+    # Test dict return value
+    def evaluate_config_dict(config):
+        metric = (round(config['x'])-85000)**2 - config['x']/config['y']
+        return {"metric": metric}
+
+    tune.run(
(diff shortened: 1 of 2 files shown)


eduardobull:
Resolves #142

sonichi (commented):
Thanks. Could you add your test that covers this case in test/tune/test_tune.py?

test/tune/test_tune.py
```diff
@@ -217,6 +217,22 @@ def simple_func(config):
...
 
+def test_run_training_function_return_value():
+    from flaml import tune
+
+    def evaluate_config(config):
+        metric = (round(config['x'])-85000)**2 - config['x']/config['y']
+        return {"metric": metric}
+
+    tune.run(
+        evaluate_config,
```
sonichi, line 228: The test failed. Either pass `metric='metric'` to tune.run or return a float in `evaluate_config`

sonichi (changes requested):
Please sign

test/tune/test_tune.py
eduardobull: I've modified the test function to pass a `training_function` returning a dict and another one returning a scalar.

sonichi (approved):

PR #166, opened 2021-08-15
fix typo

qingyun-wu:


sonichi (commented):
How about merging with branch v0.5.13?

PR #170, opened 2021-08-19
support ROC and AUC for multi-class classification

su2umaru:


su2umaru:
I've challenged to create a fix. Could you please have a moment to take a look at it?

flaml/ml.py
```diff
@@ -64,6 +65,7 @@ def sklearn_metric_loss_score(
         y_true: A 1d numpy array of the true labels
         labels: A 1d numpy array of the unique labels
         sample_weight: A 1d numpy array of the sample weight
+        multi_class: 'ovr' (one-vs-rest) or 'ovo' (one-vs-one)
```
sonichi, line 68: The default value is set to None. Could you document the expected behavior when it's None?

sonichi (commented):
Could you add/modify a test case in test/test_automl.py to cover the newly added metric? Actually, you also need to modify the code which calls `sklearn_metric_loss_score` and pass the right `multi_class`.

su2umaru:
I've added a test case in test/test_automl.py and modified the code which calls `sklearn_metric_loss_score`.

su2umaru:
Could you please check again?

sonichi:
Have you tested locally? Please refer to https://github.com/microsoft/FLAML#developing and make sure your test passes.

test/test_automl.py
```diff
@@ -344,6 +345,32 @@ def test_micro_macro_f1(self):
...
+            "log_training_metric": True,
+            "n_jobs": 1,
+            "model_history": True
+        }
+        X_train, y_train = load_breast_cancer(return_X_y=True)
+        automl_experiment_binary.fit(
+            X_train=X_train, y_train=y_train, **automl_settings)
+
+        automl_experiment_multi = AutoML()
+        automl_settings["multi_class"] = "ovr"
```
sonichi, line 364: I think there will be an error when you pass `multi_class="ovr"` to the `fit()` of the underlying estimators, such as LGBMClassifier.
To avoid that error, an alternative way I can suggest is to use 'roc_auc_ovr' and 'roc_auc_ovo' as the new metric name and remove the `multi_class` argument from `fit_kwargs`.

su2umaru:
I've made sure my test passes. Could you please review it again?

sonichi (approved):

PR #184, opened 2021-09-02
Fix decide_split_type bug.

gianpDomiziani:
I’ve noted a small bug in the decide_split_type method of the automl module. The ‘binary’ problem was not included.

### DIFF
--- a/flaml/automl.py
+++ b/flaml/automl.py
@@ -911,7 +911,7 @@ class AutoML:
         return training_duration
 
     def _decide_split_type(self, split_type):
-        if self._state.task == 'classification' or self._state.task == 'binary':
+        if self._state.task in ('classification', 'binary', 'multi'):
             self._state.task = get_classification_objective(
                 len(np.unique(self._y_train_all)))
             assert split_type in [None, "stratified", "uniform", "time"]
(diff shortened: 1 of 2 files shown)


flaml/automl.py
```diff
@@ -911,7 +911,7 @@ def retrain_from_log(self,
         return training_duration
 
     def _decide_split_type(self, split_type):
-        if self._state.task == 'classification':
+        if self._state.task == 'classification' or self._state.task == 'binary':
```
sonichi, line 914:
```suggestion
        if self._state.task in ('classification', 'binary', 'multi'):
```

test/test_automl.py
```diff
@@ -282,6 +282,20 @@ def test_custom_metric(self):
                 filename=automl_settings['log_file_name'], time_budget=6)
         print(metric_history)
 
+    def test_binary(self):
+        automl_experiment = AutoML()
+        automl_settings = {
+            "time_budget": 4,
```
sonichi, line 288:
```suggestion
            "time_budget": 1,
```

sonichi (changes requested):

sonichi (approved):

PR #186, opened 2021-09-02
warmstart blendsearch

sonichi:
Support warmstart with evaluated rewards
Increase test coverage
Only use define-by-run for hierarchical search space


flaml/tune/sample.py
```diff
@@ -344,55 +339,6 @@ def domain_str(self):
         return f"{self.categories}"
 
 
-class Function(Domain):
```
qingyun-wu, line 347: why removing this?

test/test_training_log.py
```diff
@@ -46,6 +46,7 @@ def test_training_log(self, path='test_training_log.log'):
             automl_settings["log_file_name"] = None
             automl_experiment.fit(X_train=X_train, y_train=y_train,
                                   **automl_settings)
+            automl_experiment._selected.update(None, 0)
```
qingyun-wu, line 49: This line is added intentionally to test the update function of SearchState? Does not look good.

sonichi: It's for improving the coverage. Any alternative way to improve coverage?

flaml/tune/sample.py
sonichi: not tested and not used anywhere.

flaml/searcher/suggestion.py
```diff
@@ -152,83 +143,6 @@ def on_trial_result(self, trial_id: str, result: Dict):
         """
         pass
 
-    def on_trial_complete(self,
```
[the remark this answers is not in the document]
sonichi: not tested and not used anywhere.

flaml/scheduler/trial_scheduler.py
```diff
@@ -31,127 +29,10 @@ class TrialScheduler:
...
-        raise NotImplementedError
-
-    def on_trial_error(self, trial_runner: "trial_runner.TrialRunner",
-                       trial: Trial):
-        """Notification for the error of trial.
-        This will only be called when the trial is in the RUNNING state."""
-
-        raise NotImplementedError
-
-    def on_trial_result(self, trial_runner: "trial_runner.TrialRunner",
```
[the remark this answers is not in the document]
sonichi: It doesn't affect OnlineScheduler.

flaml/scheduler/trial_scheduler.py
```diff
@@ -31,127 +29,10 @@ class TrialScheduler:
...
-
-    def on_trial_remove(self, trial_runner: "trial_runner.TrialRunner",
-                        trial: Trial):
-        """Called to remove trial.
-        This is called when the trial is in PAUSED or PENDING state. Otherwise,
-        call `on_trial_complete`."""
-
-        raise NotImplementedError
-
-    def choose_trial_to_run(
```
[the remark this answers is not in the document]
sonichi: It doesn't affect OnlineScheduler.

test/test_training_log.py
sonichi: Remember this is testing code.

qingyun-wu (approved):

flaml/automl.py
```diff
@@ -456,30 +451,29 @@ def _validate_data(self, X_train_all, y_train_all, dataframe, label,
...
-                raise ValueError(
-                    "X_train_all must be a numpy array, a pandas dataframe, "
-                    "or Scipy sparse matrix.")
-            if not (isinstance(y_train_all, np.ndarray)
-                    or isinstance(y_train_all, pd.Series)):
-                raise ValueError(
-                    "y_train_all must be a numpy array or a pandas series.")
-            if X_train_all.size == 0 or y_train_all.size == 0:
-                raise ValueError("Input data must not be empty.")
+            assert (
```
ekzhu, line 454: Generally `assert` is reserved for debugging or panic that must shutdown everything (see this [post](https://stackoverflow.com/questions/944592/best-practice-for-using-assert)). I think the old style is better -- it allows the user's program to catch and fix the error.

ekzhu (commented):
Don't forget to update version number.

flaml/automl.py
ekzhu: An easy way to get around this is to replace `assert` with a custom function `check_value`, which raises a ValueError if the condition is false.

ekzhu (approved):

flaml/automl.py
sonichi: Agree to have a better approach in future. For the record, using assert is not totally unreasonable: https://code-examples.net/en/q/e69d0.

PR #191, opened 2021-09-07
V0.6.3

qingyun-wu:
- main/v0.6.3
- extend starting_points: can provide a list of starting points for each estimator in the starting_points dict


flaml/automl.py
```diff
@@ -1249,6 +1255,8 @@ def custom_metric(
                 config for the estimators.
                 Keys are the name of the estimators, and values are the starting
                 hyperparamter configurations for the corresponding estimators.
+                The value can be a single hyperparamter configuration dict or a list
+                of hyperparamter configuration dict.
```
sonichi, line 1259:
```suggestion
                of hyperparamter configuration dicts.
```

PR #282, opened 2021-11-14
add max_depth to xgboost search space

sonichi:
For improving performance on regression.
Example dataset: brazil_houses.


flaml/model.py
```diff
@@ -525,6 +525,10 @@ def search_space(cls, data_size, **params):
                 "init_value": 4,
                 "low_cost_init_value": 4,
             },
+            "max_depth": {
```
qingyun-wu, line 528: Is it necessary to add `max_depth` to the search_space here? `max_depth` is removed from the space in XGBoostSklearnEstimator, and added in XGBoostLimitDepthEstimator 
```python
space["max_depth"] = {
            "domain": tune.randint(lower=1, upper=min(upper, 16)),
            "init_value": 1,
            "low_cost_init_value": 1,
        }
```

sonichi: XGBoostEstimator is not used as a default estimator. It's used for a custom estimator example. It's not necessary to add it here. I added it just to keep an example of including both 0 and non-zero in the domain of max_depth.

qingyun-wu (approved):

PR #306, opened 2021-11-24
Update test_regression.py

fengsxy:
There is a another way for mutioutput-model-trian.
RegressorChain is more  adapting to the targets which are relavant.

### DIFF
--- a/test/automl/test_regression.py
+++ b/test/automl/test_regression.py
@@ -198,7 +198,7 @@ class TestRegression(unittest.TestCase):
 def test_multioutput():
     from sklearn.datasets import make_regression
     from sklearn.model_selection import train_test_split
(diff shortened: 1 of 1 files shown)


test/automl/test_regression.py
```diff
@@ -198,7 +198,7 @@ def test_regression_xgboost(self):
 def test_multioutput():
     from sklearn.datasets import make_regression
     from sklearn.model_selection import train_test_split
-    from sklearn.multioutput import MultiOutputRegressor
+    from sklearn.multioutput import MultiOutputRegressor,RegressorChain
```
sonichi, line 201:
```suggestion
    from sklearn.multioutput import MultiOutputRegressor, RegressorChain
```

sonichi (approved):

PR #319, opened 2021-11-30
pred_time_limit clarification and logging

### LINKED ISSUE #318: What exactly does pred_time_limit refer to?
`pred_time_limit` sounds like a great way to limit complexity of models being fitted; but what time does it refer to: inference on one row, inference on the whole training dataset, or something else?

sonichi:
resolve #318


EgorKraevTransferwise (commented):
That certainly answers my original question, so is an improvement on the status quo, but doesn't explain how to choose reasonable values - is the pred_time of models that are being fit as part of the AutoML process being logged somewhere, so one could retrieve it and look at the range of values it takes?
This could be a super useful argument, if only I understood how to use it well :)

sonichi:
> That certainly answers my original question, so is an improvement on the status quo, but doesn't explain how to choose reasonable values - is the pred_time of models that are being fit as part of the AutoML process being logged somewhere, so one could retrieve it and look at the range of values it takes? This could be a super useful argument, if only I understood how to use it well :)

Thanks for the suggestion. I just updated the PR to log the prediction time in the log file. Please let me know if that addresses what you asked.

qingyun-wu (approved):

EgorKraevTransferwise:
Perfect, thanks a lot!

PR #323, opened 2021-12-04
bug fix in confg2params

sonichi:
* rename "cls" -> "self"
* pop "criterion" for rf regression


sonichi:
* rename "cls" -> "self"
* pop "criterion" for rf regression

qingyun-wu (approved):

PR #333, opened 2021-12-08
support for customized splitters

wuchihsu:
A customized splitter can be passed via `split_type`. It can be an iterable or an object that has `split` and `get_n_splits` methods.


flaml/ml.py
```diff
@@ -319,8 +319,11 @@ def evaluate_model_CV(
         shuffle = False
     elif isinstance(kf, TimeSeriesSplit):
         kf = kf.split(X_train_split, y_train_split)
-    else:
+    elif hasattr(kf, "split"):
         kf = kf.split(X_train_split)
+    else:
+        n = len(kf)
```
slhuang, line 325: any constraint on the iterable object `split_type` in automl.py, such that kf is already in `train_index, test_index` ndarray?

flaml/automl.py
```diff
@@ -507,7 +509,8 @@ def custom_metric(
                 True - retrain only after search finishes; False - no retraining;
                 'budget' - do best effort to retrain without violating the time
                 budget.
-            split_type: str, default="auto" | the data split type.
+            split_type: str or iterable, object that has split and get_n_splits methods,
```
slhuang, line 512: what is the use case of having split_type in iterable type? Maybe elaborate a bit or give an example.

slhuang: Please add a test in test/automl/test_split.py for both cases: iterable, and class with split/get_n_splits methods.

flaml/ml.py
```diff
@@ -319,8 +319,11 @@ def evaluate_model_CV(
         shuffle = False
     elif isinstance(kf, TimeSeriesSplit):
         kf = kf.split(X_train_split, y_train_split)
-    else:
+    elif hasattr(kf, "split"):
         kf = kf.split(X_train_split)
+    else:
```
slhuang, line 324: Use `elif isinstance(kf, list)`. Keep `kf = kf.split(X_train_split)` as default, as there might be other split_type.

flaml/ml.py
```diff
@@ -298,7 +298,7 @@ def evaluate_model_CV(
     metric = None
     train_time = pred_time = 0
     valid_fold_num = total_fold_num = 0
-    n = kf.get_n_splits()
+    n = kf.get_n_splits() if hasattr(kf, "get_n_splits") else None
```
slhuang, line 301: else len(kf)?

flaml/automl.py
slhuang: We may need to elaborate what are the constraints for `split_type` when it is iterable, or object -- what is input/output for `split()` and `get_n_split()`.

Do you think it would be easier if we have one single API for customized splitter: users implement a derived class of https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.KFold.html#sklearn.model_selection.KFold, customizing split() and get_n_splits() function but with the same signature. `Iterable split_type` can be implemented using this API as well. This way, we have a more concise API and do not need to worry about whether the input `iterable split_type` is in correct form or not.

wuchihsu: > We may need to elaborate what are the constraints for `split_type` when it is iterable, or object -- what is input/output for `split()` and `get_n_split()`.
> 
> Do you think it would be easier if we have one single API for customized splitter: users implement a derived class of https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.KFold.html#sklearn.model_selection.KFold, customizing split() and get_n_splits() function but with the same signature. `Iterable split_type` can be implemented using this API as well. This way, we have a more concise API and do not need to worry about whether the input `iterable split_type` is in correct form or not.

Yeah, the definitions of an `iterable`, `split()` and `get_n_split()` are not very clear. It is clearer to use a derived class of `KFold`. I can try to change it to that. Do I need to open a new PR then.

slhuang: Is it hard to check the signature during runtime? If so, maybe we can just write down the API signature of `split()` and `get_n_split()` in example/comments/docs and catch the exception during runtime and report the error?

wuchihsu: OK, I can try this

flaml/automl.py
```diff
@@ -507,7 +507,11 @@ def custom_metric(
                 True - retrain only after search finishes; False - no retraining;
                 'budget' - do best effort to retrain without violating the time
                 budget.
-            split_type: str, default="auto" | the data split type.
+            split_type: str or splitter object, default="auto" | the data split type.
+                Object must be a derived class of scikit-learn KFold
+                (https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.KFold.html#sklearn.model_selection.KFold)
+                and have ``split`` and ``get_n_splits`` methods with the same signatures.
```
slhuang, line 513: Thanks. These all look good to me. 
One other thing we need to handle when calling split() in ml.py:  https://github.com/microsoft/FLAML/blob/dd60dbc5eb7e6d298c5f1506a8e0b7a8bb390052/flaml/ml.py#L319
When task == TS_FORCAST, always set shuffle=False to avoid data shuffling. 
The simplest way is to set this line 
https://github.com/microsoft/FLAML/blob/dd60dbc5eb7e6d298c5f1506a8e0b7a8bb390052/flaml/ml.py#L308
`shuffle=False if task == TS_FORECAST else True` and update https://github.com/microsoft/FLAML/blob/dd60dbc5eb7e6d298c5f1506a8e0b7a8bb390052/flaml/ml.py#L315 and https://github.com/microsoft/FLAML/blob/dd60dbc5eb7e6d298c5f1506a8e0b7a8bb390052/flaml/ml.py#L319 accordingly.

wuchihsu: If `shuffle = False if task == TS_FORECAST else True` at Line 308, not quite sure what needs to be updated at Line 315 and 319. What if they remain the same?

slhuang: please just remove line 319 after changing line 308, as 319 is already handled by 308.

slhuang (approved):

flaml/automl.py
```diff
@@ -1277,7 +1284,11 @@ def retrain_from_log(
                 ['auto', 'cv', 'holdout'].
             split_ratio: A float of the validation data percentage for holdout.
             n_splits: An integer of the number of folds for cross-validation.
-            split_type: str, default="auto" | the data split type.
+            split_type: str or splitter object, default="auto" | the data split type.
+                Object must be a derived class of scikit-learn KFold
```
sonichi, line 1288:
```suggestion
                A valid splitter object is an instance of a derived class of scikit-learn KFold
```

flaml/automl.py
```diff
@@ -1399,7 +1410,12 @@ def _decide_split_type(self, split_type):
             self._state.task = get_classification_objective(
                 len(np.unique(self._y_train_all))
             )
-        if self._state.task in CLASSIFICATION:
+        if not isinstance(split_type, str):
+            assert hasattr(split_type, "split") and hasattr(
+                split_type, "get_n_splits"
+            ), "split_type must be a string or splitter object with split and get_n_splits methods."
```
sonichi, line 1416:
```suggestion
            ), "split_type must be a string or a splitter object with split and get_n_splits methods."
```

flaml/ml.py
```diff
@@ -423,7 +423,6 @@ def evaluate_model_CV(
         y_train_all = pd.DataFrame(y_train_all, columns=[TS_VALUE_COL])
         train = X_train_all.join(y_train_all)
         kf = kf.split(train)
```
sonichi, line 425: Can we remove line 422 to 425? @int-chaos

flaml/automl.py
```diff
@@ -955,7 +959,7 @@ def _prepare_data(self, eval_method, split_ratio, n_splits):
             self._state.task in CLASSIFICATION
             and self._auto_augment
             and self._state.fit_kwargs.get("sample_weight") is None
-            and self._split_type not in ["time", "group"]
+            and self._split_type in ["auto", "stratified", "uniform"]
```
sonichi, line 962:
```suggestion
            and self._split_type in ["stratified", "uniform"]
```

flaml/automl.py
```diff
@@ -507,7 +507,11 @@ def custom_metric(
                 True - retrain only after search finishes; False - no retraining;
                 'budget' - do best effort to retrain without violating the time
                 budget.
-            split_type: str, default="auto" | the data split type.
+            split_type: str or splitter object, default="auto" | the data split type.
```
sonichi, line 510: Please update the docstr for split_type in `__init__()` and `retrain_from_log()`

wuchihsu: OK, also in `fit()`.

sonichi (approved):

PR #358, opened 2021-12-24
Simplify lgbm example

ruizhuanguw:
lightgbm.train is able to take num_boost_round information from config. There is no need to seperate num_boost_round from other configuration parameters.

### DIFF
--- a/website/docs/Getting-Started.md
+++ b/website/docs/Getting-Started.md
@@ -74,7 +74,7 @@ analysis = tune.run(
     low_cost_partial_config=low_cost_partial_config, time_budget_s=3, num_samples=-1,
 )
 ```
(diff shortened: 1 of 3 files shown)


website/docs/Getting-Started.md
```diff
@@ -75,6 +74,7 @@ analysis = tune.run(
     low_cost_partial_config=low_cost_partial_config, time_budget_s=3, num_samples=-1,
 )
 ```
+Please see [script](https://github.com/microsoft/FLAML/blob/main/test/tune.py) for the complete version of the above example.
```
sonichi, line 77: shall we update that script as well?
```suggestion
Please see this [script](https://github.com/microsoft/FLAML/blob/main/test/tune.py) for the complete version of the above example.
```

sonichi (approved):

PR #388, opened 2022-01-09
fix issues in logging, bug in space.py, constraint sign, and improve code coverage

sonichi:
* add console log handler under mlflow run.
* comment on skippable steps in website test. @qingyun-wu
* sign for config and metric constraints. @dtquandt 
* bug fix for define-by-run and unflatten_hierarchical.
* test for code coverage. @qingyun-wu the coverage for onlineml is lagging behind.
* handle nested space in indexof().
* always increase iter count regardless of sample size. @liususan091219 
* security patch.
* handle checkpoints properly: not relying on custom metric any more. Both standard metric and custom metric can log checkpoint results. Revised the custom metric test as well. @liususan091219


liususan091219 (approved):

flaml/automl.py
```diff
@@ -2326,20 +2326,20 @@ def _search_parallel(self):
                 wall_time = result.get("wall_clock_time")
                 if wall_time is not None:
                     self._state.time_from_start = wall_time
+                self._iter_per_learner[estimator] += 1
```
qingyun-wu, line 2329: Do we also need to change line 2561 (the sequential case)?

flaml/automl.py
```diff
@@ -2315,7 +2315,7 @@ def _search_parallel(self):
             ),
             key=lambda x: x.last_result["wall_clock_time"],
         )
-        for _track_iter, trial in enumerate(trials):
+        for self._track_iter, trial in enumerate(trials):
```
qingyun-wu, line 2318: This makes line 571 redundant?

flaml/automl.py
sonichi: line 571 is needed for search_sequential()

flaml/automl.py
sonichi: line 571 is needed for search_sequential()

flaml/automl.py
sonichi: In the sequential case, we use this field in line 2852, so we can't change line 2561. Maybe I should use a different variable to avoid the inconsistency.

qingyun-wu (approved):

PR #421, opened 2022-01-22
Use Ray `_BackwardsCompatibleNumpyRng` if possible

Yard1:
Fixes CI issues on Ray side that prevent us from testing the latest FLAML versions (see https://buildkite.com/ray-project/ray-builders-pr/builds/22971#03fbedc2-8a2d-4943-8c2b-efb5f92ca0fd)


sonichi (approved):
Looks good. Is the new random sampler released in ray 1.9.2?

Yard1:
@sonichi It will be released in 1.10, but the change can be made right now

PR #430, opened 2022-01-25
data in csv

sonichi:
* Use data in csv file in the UDF example.
* Use object store to store the state including the data, when using ray.
* Added tests and examples using object store.


flaml/automl.py
```diff
@@ -1667,9 +1667,18 @@ def trainable(self) -> Callable[[dict], Optional[float]]:
...
-                search_state.training_function = partial(
-                    AutoMLState._compute_with_config_base, self._state, estimator
-                )
+                if self._use_ray:
+                    from ray.tune import with_parameters
+
+                    search_state.training_function = with_parameters(
+                        AutoMLState._compute_with_config_base,
+                        self=self._state,
+                        estimato=estimator,
```
camer314, line 1676: This looks like a typo

camer314 (approved):

torronen:
About sample:
 I receive  "lightgbm.basic.LightGBMError: Input format error when parsing as CSV"
 I have run CSV validation with csvkit, and the data only includes numbers.

I will try to find the solution and what is the expected CSV format for LightGBM

It may not be ideal for general sample, because not even ignored cols are allowed to have text:
https://github.com/microsoft/LightGBM/issues/1840

test/tune_example.py
```diff
@@ -5,17 +5,27 @@
...
+# )
 
 
 def train_lgbm(config: dict) -> dict:
     # convert config dict to lgbm params
     params = LGBMEstimator(**config).params
     # train the model
-    train_set = lightgbm.Dataset(X_train, y_train)
+    # train_set = lightgbm.Dataset(X_train, y_train)
+    train_set = lightgbm.Dataset(
```
torronen, line 26: If here the does it cause the dataset to be loaded for each iteration? Suggest: move outside train_lgbm function

torronen: Can not: lightgbm.basic.LightGBMError: Cannot change max_bin after constructed Dataset handle.

torronen (approved):

torronen:
BTW, I receive this warning
UserWarning: Found `n_estimators` in params. Will use it instead of argument
  _log_warning(f"Found `{alias}` in params. Will use it instead of argument")
  
  but I can not figure out why.

test/tune_example.py
```diff
@@ -55,3 +65,4 @@ def train_lgbm(config: dict) -> dict:
     time_budget_s=3,
     num_samples=-1,
 )
+print(analysis.best_result)
```
torronen, line 68: print(analysis.best_config) ?

analysis.best_results throws _`AttributeError: 'ExperimentAnalysis' object has no attribute 'best_result'`_ for me

test/tune_example.py
sonichi: Right, the actual loading happens in the training.

test/tune_example.py
```diff
@@ -5,17 +5,27 @@
...
     params = LGBMEstimator(**config).params
     # train the model
-    train_set = lightgbm.Dataset(X_train, y_train)
+    # train_set = lightgbm.Dataset(X_train, y_train)
+    train_set = lightgbm.Dataset(
+        csv_file_name, params={"label_column": "name:MedHouseVal", "header": True}
+    )
     model = lightgbm.train(params, train_set)
     # evaluate the model
     pred = model.predict(X_test)
```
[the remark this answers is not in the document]
sonichi: Feel free to create another PR after you figure that out.

test/tune_example.py
sonichi: Thanks. I added the `best_result` property.

qingyun-wu (approved):

PR #436, opened 2022-01-28
Change the upper bound for "lags" hyperparameter for sklearn forecast models

int-chaos:
- Change the upper bound for "lags" hyperparameter for sklearn forecast models
- seek to fix memory error and debug errors with lags


int-chaos:
Has commit history from a previous pr. Closing PR. Refer to new PR.

Issue #424, opened 2022-01-23
code crash jupyter kernel


Tint0ri:
try to run a s simple ts forecast code on Google Colab(12G RAM) and Amzon Sagemaker(16G RAM). Both environment get jupyter kernel crashed and restarted report after a while. 

sample code:

import numpy as np
import pandas as pd
from flaml import AutoML

X_train = pd.date_range('2017-01-01', periods=70000, freq='T')
y_train = pd.DataFrame(np.random.randint(6500, 7500, 70000))
automl = AutoML()
automl.fit(X_train=X_train[:-10].values,  # a single column of timestamp
           y_train=y_train[:-10].values,  # value for each timestamp
           period=10,  # time horizon to forecast, e.g., 12 months
           task='ts_forecast', time_budget=8000,  # time budget in seconds
          )

No error output but kernel restarted and training process stopped. Maybe caused by memory management issue.

sonichi:
I believe the root cause is the same as #412 . I suggest the same workaround before the issue is fixed.
@int-chaos

Tint0ri:
I had the #412 issue before. But I can bypass it by change period from 12 to a smaller number, such as 10 as above. My code can finish if time_budget is a small number.  But if I set time_budget greater than 2000, I got kernel crashed. I assumed #424 should be something about memory management such as not retrieve the memory in time. Because Colab log show memory tmalloc was not fulfilled.

int-chaos:
To address this memory issue, we will be setting a limit for the upper bound of `lags`. But what exactly should be the limit is to be determined. If you have any suggestions for it please let me know. Below is an explanation of what is happening and why this error occurs.

Currently our upper bound for the `lags` hyperparameter is dependent on the size of the data_size and period, where the upper bound is equal to data_size - period. It seems like you have a really large training set with a data size of 69990 timestamps and you're forecasting a period of 10. With a split of 5, this means that the upper bounds of `lags` are 69970, 69960, 69950, 69940 and 69930 for your five validation sets. When you increase the time_budget to be greater than 2000 it will explore a `lag` value that approaches the upper limit. It seems like it gets to a point where the number of lag features is too great for the memory.

sonichi:
And that's why the same workaround in https://github.com/microsoft/FLAML/issues/412#issuecomment-1014920806 would avoid this issue.

int-chaos:
[...]

Issue #412, opened 2022-01-17
Forecasting example


Yasslight90:
Hello,

The example on https://github.com/microsoft/FLAML for forecasting is not working with the latest release 0.9.5. Please find the error below:

InsufficientDataLengthError: Sum of model lags (3) and forecasting horizon (12 cannot be bigger than length of y (12).

Kind regards,

Yassine

sonichi:
It looks a bug for the newly added regressors. @int-chaos 

As a workaround, you can limit the estimator_list to `["prophet", "arima", "sarimax"]`.

Yasslight90:
I just came back to and old release. Let me know when you release a new version. 
Thank you!

int-chaos:
Thank you for the issue. I'll look into it more, but it seems the error occurs because the lag value is too big.

sonichi:
> I just came back to and old release. Let me know when you release a new version. Thank you!

@Yasslight90 v0.9.6 is released.

Yasslight90:
Ok perfect. I will test it.

PR #437, opened 2022-01-28
Change the upper bound for "lags" hyperparameter for sklearn forecast models

int-chaos:
- Change the upper bound for "lags" hyperparameter for sklearn forecast models
- seek to fix memory error and debug errors with lags

#424 #412

### DIFF
--- a/flaml/model.py
+++ b/flaml/model.py
@@ -1814,7 +1814,8 @@ class TS_SKLearn(SKLearnEstimator):
                 },
                 "lags": {
                     "domain": tune.randint(
-                        lower=1, upper=np.floor(np.sqrt(data_size[0]))
+                        lower=1, upper=int(np.sqrt(data_size[0]))
+
                     ),
                     "init_value": 3,
                 },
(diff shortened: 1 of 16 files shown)


int-chaos:
ran `test_numpy_large()` in test_forecast.py with `time_budget=3000` locally and it worked with causing a memory issue

flaml/model.py
```diff
@@ -1813,7 +1813,9 @@ def search_space(cls, data_size, pred_horizon, **params):
                     "low_cost_init_value": False,
                 },
                 "lags": {
-                    "domain": tune.randint(lower=1, upper=data_size[0] - pred_horizon),
+                    "domain": tune.randint(
+                        lower=1, upper=np.floor(np.sqrt(data_size[0]))
```
sonichi, line 1817:
```suggestion
                        lower=1, upper=int(np.sqrt(data_size[0]))

```

sonichi (commented):
Need to update the example's output in https://microsoft.github.io/FLAML/docs/Examples/AutoML-Time%20series%20forecast

sonichi (approved):

PR #455, opened 2022-02-11
fix a bug when using ray & update ray on aml

sonichi:


Yard1:
@sonichi What was the bug with Ray?

sonichi:
> @sonichi What was the bug with Ray?

When using with_parameters(), the config argument must be the first argument in the trainable function. It wasn't in automl.py.

Yard1 (approved):

camer314 (approved):

rezasherafat (approved):

Issue #504, opened 2022-03-26
How to report a failed evaluation of evaluation_function to tune?


EgorKraevTransferwise:
I am trying to optimize a custom objective function, for which some evaluations may fail with an exception. I can catch the exceptions, but what is a graceful return value for the evaluation function to return to `tune.run` if the evaluation of that particular config has failed?

I could just return a very bad value of the metric being optimized, but wondering whether there is a better way, and couldn't find it in the docs.

Thanks a lot!

sonichi:
Does the failure mean the configuration is bad? If so, returning a bad value of the metric is right.

EgorKraevTransferwise:
We search over ML models from the EconML repo (and their hyperparameters), and as some models there are fairly fresh, certain models can fail on certain datasets for all sorts of sometimes unforeseen reasons; the goal is to handle that gracefully and keep searching over models/configurations that work

EgorKraevTransferwise:
So do I understand correctly that with #505 merged, the evaluation_function should return an empty dict in this case? Would it not make sense to return a metric value of NaN instead? That way, the evaluation function could return other data in the return dict, eg details of the failure.

sonichi:
In your scenario, I still suggest returning a bad value of the metric being optimized because the failure does imply the configuration is bad for your dataset. The empty dict is meant for undeterministic evaluation error which doesn't imply the configuration is bad.

EgorKraevTransferwise:
But for "bad" I'd have to guess ahead of time the likely range of the metric and then return a valid float value worse than anything in that range, right? There is no support for eg a NaN metric value to indicate such a usecase?

sonichi:
> But for "bad" I'd have to guess ahead of time the likely range of the metric and then return a valid float value worse than anything in that range, right? There is no support for eg a NaN metric value to indicate such a usecase?

You can return np.inf or -np.inf depending on whether you minimize or maximize the metric.

EgorKraevTransferwise:
Perfect, thanks! Will try that.

PR #505, opened 2022-03-26
handle failing trials

sonichi:
When a trial fails unexpectedly and we don't necessarily know the configuration is bad, allow the training function to return an empty dict when not using ray to inform the tuner about the error.
When using ray, the training function can raise an error (@Yard1 could you please confirm that).
#504


Yard1:
If an exception is raised in the trainable, the trial will be marked as errored by Ray. Returning an empty dict will also raise an error due to inconsistent keys being reported, but that can be suppressed with an env var.

qingyun-wu (approved):

PR #532, opened 2022-05-01
handle non-flaml scheduler in flaml.tune

qingyun-wu:


flaml/tune/tune.py
```diff
@@ -109,7 +109,7 @@ def compute_with_config(config):
         if _verbose > 2:
             logger.info(f"result: {result}")
         if trial.is_finished():
-            return None
+            return False
         else:
             return True
```
sonichi, line 114: return not trial.is_finished()

sonichi: add docstr about returns.

flaml/tune/tune.py
```diff
@@ -236,9 +236,21 @@ def compute_with_config(config):
...
             and how to set up the corresponding evaluation functions in
-            test/tune/test_scheduler.py. TODO: point to notebook examples.
+            test/tune/test_scheduler.py, and test/tune/example_scheduler.py.
+    ```python
+    def easy_objective(config):
+        width, height = config["width"], config["height"]
+        for step in range(config["steps"]):
+            intermediate_score = evaluation_fn(step, width, height)
+            trial_not_finished = tune.report(iterations=step, mean_loss=intermediate_score)
+            if trial_not_finished is False:
```
sonichi, line 251:
```suggestion
            keep_running = tune.report(iterations=step, mean_loss=intermediate_score)
            if keep_running is False:
```

flaml/tune/tune.py
```diff
@@ -236,9 +236,21 @@ def compute_with_config(config):
...
+                return
```
sonichi, line 252: I'm not sure if we should return None or return the intermediate_score. Could you check?

flaml/tune/tune.py
```diff
@@ -388,7 +400,11 @@ def compute_with_config(config):
             searcher.set_search_properties(metric, mode, config, setting)
         else:
             searcher.set_search_properties(metric, mode, config)
-    if scheduler == "asha":
+    if (
+        scheduler == "asha"
+        or scheduler == "asynchyperband"
+        or scheduler == "async_hyperband"
+    ):
```
sonichi, line 407:
```suggestion
    if scheduler in ("asha", "asynchyperband", "async_hyperband"):
```

test/tune/example_scheduler.py
```diff
@@ -0,0 +1,106 @@
...
+        trial_not_finished = tune.report(iterations=step, mean_loss=intermediate_score)
+        if trial_not_finished is False:
+            return
+
+
+def test_tune_scheduler(smoke_test=True, use_ray=True, use_raytune=False):
+    import numpy as np
+    from flaml.searcher.blendsearch import BlendSearch
+
+    np.random.seed(100)
```
sonichi, line 30: what for?

website/docs/Use-Cases/Tune-User-Defined-Function.md
```diff
@@ -350,7 +350,9 @@ tune.run(.., scheduler=my_scheduler, ...)
...
+- Different from the case when the `flaml` scheduler is used, the amount of resources to use at each iteration is not suggested by the search algorithm through the `resource_attr` in a configuration. You need to specify the evaluation schedule explicitly by yourself in the `evaluation_function` and **report intermediate results (using `tune.report()`) accordingly**. In the following code example, we use the ASHA scheduler by setting `scheduler="asha"`, we specify `resource_attr`, `min_resource`, `min_resource` and `reduction_factor` the same way as in the previous example (when "flaml" is used as the scheduler). We perform the evaluation in a customized schedule.
```
sonichi, line 353:
```suggestion
- Different from the case when the `flaml` scheduler is used, the amount of resources to use at each iteration is not suggested by the search algorithm through the `resource_attr` in a configuration. You need to specify the evaluation schedule explicitly by yourself in the `evaluation_function` and **report intermediate results (using `tune.report()`) accordingly**. In the following code example, we use the ASHA scheduler by setting `scheduler="asha"`. We specify `resource_attr`, `min_resource`, `min_resource` and `reduction_factor` the same way as in the previous example (when "flaml" is used as the scheduler). We perform the evaluation in a customized schedule.
```

website/docs/Use-Cases/Tune-User-Defined-Function.md
```diff
@@ -350,7 +350,9 @@ tune.run(.., scheduler=my_scheduler, ...)
...
+
+- Use ray backend or not? You can choose to use ray backend or not by specifying `use_ray=True` or  `use_ray=False`. When ray backend is not used, i.e., `use_ray=False`, you need to stop the evaluation function according to the returned result of `tune.report()` as shown in the last two lines of the evaluation function`obj_w_intermediate_report()` in the following code example.
```
sonichi, line 355:
```suggestion
- Use ray backend or not? You can choose to use ray backend or not by specifying `use_ray=True` or  `use_ray=False`. When ray backend is not used, i.e., `use_ray=False`, you need to stop the evaluation function according to the returned result of `tune.report()` as shown in the last two lines of the evaluation function `obj_w_intermediate_report()` in the following code example.
```

sonichi (commented):
How's asha performing now in your example?

test/tune/example_scheduler.py
qingyun-wu: Was trying to check the consistency between 'flaml.tune' and 'ray.tune'

flaml/tune/tune.py
```diff
@@ -78,6 +78,9 @@ def compute_with_config(config):
         _metric: Optional default anonymous metric for ``tune.report(value)``.
             (For compatibility with ray.tune.report)
         **kwargs: Any key value pair to be reported.
+
+    Returns:
+        A bool indicating whether the trial is finished or not.
```
sonichi, line 83: This seems to indicate "True" means the trial is finished.
We can define it that way, but then need to change other places.

test/tune/example_scheduler.py
sonichi: I don't think their consistency relies on np.random.seed(). Could you confirm that?

flaml/tune/tune.py
sonichi: Another idea: What if we raise StopIteration instead of returning a bool?

flaml/tune/tune.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, trials, metric, mode):
             self.default_mode = mode
 
 
-def report(_metric=None, **kwargs):
+def report(_metric=None, **kwargs) -> bool:
```
sonichi, line 49:
```suggestion
def report(_metric=None, **kwargs):

```

flaml/tune/tune.py
```diff
@@ -78,6 +78,9 @@ def compute_with_config(config):
         _metric: Optional default anonymous metric for ``tune.report(value)``.
             (For compatibility with ray.tune.report)
         **kwargs: Any key value pair to be reported.
+
+    Raises:
+        StopIteration: A StopIteration exception is raised if the trial has been signaled to stop.
```
sonichi, line 83: only true when not using ray.

sonichi (approved):

PR #533, opened 2022-05-02
test reproducibility from retrain

sonichi:


test/automl/test_regression.py
```diff
@@ -45,37 +45,48 @@ def test_regression(self):
...
+        y_pred = automl.predict(X_train)
+        print(y_pred)
+        print(automl.model.estimator)
+        print(automl.config_history)
+        print(automl.best_model_for_estimator("xgboost"))
+        print(automl.best_iteration)
+        print(automl.best_estimator)
         print(get_output_from_log(automl_settings["log_file_name"], 1))
-        automl_experiment.retrain_from_log(
+        automl.retrain_from_log(
```
qingyun-wu, line 64: why need to keep line 64 to line 78?

qingyun-wu (approved):

Issue #476, opened 2022-03-08
A potential issue with sklearn when building an ensemble with large datasets


flippercy:
Hi @sonichi:

I noticed recently that if the dataset was large (hard to specify how 'large' it has to be), there was always a memory-related error when building the ensemble:

**TerminatedWorkerError: A worker process managed by the executor was unexpectedly terminated. This could be caused by a segmentation fault while calling the function or by an excessive memory usage causing the Operating System to kill the worker.

The exit codes of the workers are {SIGSEGV(-11), SIGSEGV(-11)}**

I did some search and it seems to be an error caused by the sklearn stacker used by flaml since people reported similar issues with sklearn.GridSearchCV (for example, https://stackoverflow.com/questions/55848101/memory-leak-using-gridsearchcv)

If I set n_jobs = 1, the issue disappeared. However, it took a lot more time to build the ensemble (not the base model).

Just want to inform you this issue. Any advice on better solutions (comparing with using just one CPU) is welcomed.

Thank you.

sonichi:
Does it fail when `n_jobs=2`?

flippercy:
I tried it with my internal datasets, default estimators from flaml and n_jobs=2; the issue still happened.

TerminatedWorkerError                     Traceback (most recent call last)
~/windows/Scorecards/Apps/R/Model-Development-Process/FLAML_Customized_Learners.py in <module>
      1 automl.fit(X_train = data_dev_balanced_B_X, y_train = data_dev_balanced_B_y,  sample_weight=data_dev_balanced_B_w,
      2           X_val = data_val_B_X, y_val = data_val_B_y, sample_weight_val=data_val_B_w,
----> 3           **settings)

~/.virtualenvs/r-reticulate/lib/python3.6/site-packages/flaml/automl.py in fit(self, X_train, y_train, dataframe, label, metric, task, n_jobs, log_file_name, estimator_list, time_budget, max_iter, sample, ensemble, eval_method, log_type, model_history, split_ratio, n_splits, log_training_metric, mem_thres, pred_time_limit, train_time_limit, X_val, y_val, sample_weight_val, groups_val, groups, verbose, retrain_full, split_type, learner_selector, hpo_method, starting_points, seed, n_concurrent_trials, keep_search_state, early_stop, append_log, auto_augment, min_sample_size, use_ray, **fit_kwargs)
   2270             with training_log_writer(log_file_name, append_log) as save_helper:
   2271                 self._training_log = save_helper
[...]

PR #551, opened 2022-05-13
choose n_jobs for ensemble according to n_jobs per learner

sonichi:
To prevent the test error in https://github.com/microsoft/FLAML/runs/6425372138?check_suite_focus=true#step:8:165
Catch the ensemble error and log the error.
Related to #476

### DIFF
--- a/flaml/automl.py
+++ b/flaml/automl.py
@@ -534,10 +534,12 @@ class AutoML(BaseEstimator):
              max_iter: An integer of the maximal number of iterations.
              sample: A boolean of whether to sample the training data during
                  search.
-             ensemble: boolean or dict | default=False. Whether to perform
-                 ensemble after search. Can be a dict with keys 'passthrough'
-                 and 'final_estimator' to specify the passthrough and
-                 final_estimator in the stacker.
+            ensemble: boolean or dict | default=False. Whether to perform
+                ensemble after search. Can be a dict with keys 'passthrough'
+                and 'final_estimator' to specify the passthrough and
+                final_estimator in the stacker. The dict can also contain
+                'n_jobs' as the key to specify the n_jobs for the stacker,
+                default to 1.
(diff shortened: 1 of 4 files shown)


Qiaochu-Song (approved):

flaml/automl.py
```diff
@@ -534,10 +534,11 @@ def custom_metric(
...
                  search.
-             ensemble: boolean or dict | default=False. Whether to perform
-                 ensemble after search. Can be a dict with keys 'passthrough'
-                 and 'final_estimator' to specify the passthrough and
-                 final_estimator in the stacker.
+            ensemble: boolean or dict | default=False. Whether to perform
+                ensemble after search. Can be a dict with keys 'passthrough'
+                and 'final_estimator' to specify the passthrough and
+                final_estimator in the stacker. The dict can also contain
+                'n_jobs' as the key to specify the n_jobs for the stacker.
```
liususan091219, line 541:
```suggestion
                'n_jobs' as the key to specify the number of jobs for the stacker.
```

flaml/automl.py
```diff
@@ -3174,18 +3183,34 @@ def _search(self):
...
+                    n_cpus = (
+                        ray.is_initialized()
+                        and ray.available_resources()["CPU"]
+                        or os.cpu_count()
+                    )
+                else:
+                    n_cpus = os.cpu_count()
+                n_jobs = (
+                    -self._state.n_jobs
+                    if abs(self._state.n_jobs) == 1
```
liususan091219, line 3198: why n_jobs = -self._state.n_jobs? This is unclear, can you add comments?

liususan091219: so when n_jobs is max, minimize it; maximize it when min? why?

liususan091219: Unresolved.

sonichi: What about now?

Issue #542, opened 2022-05-09
run test_restore.py and got _pickle.UnpicklingError: state is not a dictionary


LinWencong:
I use BlendSearch save() and restore() functions to achieve a warm start from a file, but got these problem。
so I run test_restore.py, and got the same problem 

  File "test_restore.py", line 57, in testWarmStart
    results_exp_2 = self.run_explicit_restore(r_state, checkpoint_path)
  File "test_restore.py", line 38, in run_explicit_restore
    search_alg2.restore(checkpoint_path)
  File "/usr/local/python3.6/lib/python3.6/site-packages/flaml/searcher/suggestion.py", line 252, in restore
    self.searcher.restore(checkpoint_path)
  File "/usr/local/python3.6/lib/python3.6/site-packages/flaml/searcher/blendsearch.py", line 330, in restore
    state = pickle.load(inputFile)
_pickle.UnpicklingError: state is not a dictionary

Would you mind telling me it's the way I'm using it wrong? Or is my understanding of what this function does wrong?
Thank you so much for your attention to this issue❤️

env: ubuntu 16.04  & FLAML==0.10.0

qingyun-wu:
Hi @[LinWencong](https://github.com/LinWencong),

Thank you for raising the issue and making a contribution to resolving it. I have two questions/comments:
(1) I cannot reproduce the error you reported by running test_restore.py. Can you provide more information about how you were running the test? And can you also try the latest flaml version v1.0.2 (although I did not get the error in v0.10.0 as well).
I am running the test via 
```
pytest test/tune/test_restore.py
```
using the latest code here: https://github.com/qingyun-wu/FLAML/blob/test-restore/test/tune/test_restore.py
where I added an additional test for BlendSearch (in the original file, it only has a test for CFO).

(2) The changes in your PR make `flaml.tune` and `ray.tune` not consistent which failed one of our tests: https://github.com/microsoft/FLAML/blob/main/test/tune/test_flaml_raytune_consistency.py
 
If your changes are necessary after we have a conclusion about (1), you will need to revise the PR to make `flaml.tune` and `ray.tune`. I will try to help there if the changes are necessary.

Thank you!

LinWencong:
Referring to the error message of Build / build (ubuntu-latest, 3.7), it may be wrong to simply remove the random state without saving the complete information;
But I don't understand why I can pass several of these test pipelines

LinWencong:
[...]

PR #554, opened 2022-05-18
solve issue #542. fix pickle.UnpickingError while blendsearch warm start

LinWencong:
Issue I encountered:

#542 run test_restore.py and got _pickle.UnpicklingError: state is not a dictionary

I observed:

1. numpy version
  i. When numpy==1.16*, np.random.RandomState.__getstate__() returns a tuple, not a dict.
    _pickle.UnpicklingError occurs
  ii. When numpy>1.17.0rc1, it returns a dict;
    _pickle.UnpicklingError does not occur
  iii. When numpy>1.17.0rc1, flaml uses np_random_generator = np.random.Generator,
    _pickle.UnpicklingError does not occur

2. class _BackwardsCompatibleNumpyRng
  When I remove func _BackwardsCompatibleNumpyRng.__getattr__() , _pickle.UnpicklingError doesn't occur (regardless of numpy version == 1.16* or 1.17*)

To sum up,
I think making modifications to class _BackwardsCompatibleNumpyRng is not a good choice (_BackwardsCompatibleNumpyRng came from ray）and we still need to learn more about the operation mechanism of pickle.

So I upgraded the numpy version that flaml requires：
  setup.py："NumPy>=1.17.0rc1"

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@ with open(os.path.join(here, "flaml/version.py")) as fp:
 __version__ = version["__version__"]
 
 install_requires = [
-    "NumPy>=1.16.2",
+    "NumPy>=1.17.0rc1",
     "lightgbm>=2.3.1",
     "xgboost>=0.90,<=1.3.3",
     "scipy>=1.4.1",


sonichi:
Requesting review from @skzhang1 @MichalChromcak for the version update in conda package.
Is it better to put a stable version for numpy instead of rc1?

skzhang1:
I run test_restore.py.  It runs well if upgrade numpy version. It seems no problem to me.

skzhang1 (approved):

PR #558, opened 2022-05-24
simplify warmstart in blendsearch

sonichi:
Go over the evaluated points in the beginning of search. If we don't do this, .suggest() will keep returning None for going over those points.


flaml/searcher/blendsearch.py
```diff
@@ -311,25 +311,25 @@ def _init_search(self):
...
-                    self._metric: r,
-                    self.cost_attr: 1,
-                    "config": self._evaluated_points[i],
-                }
-                for i, r in enumerate(self._all_rewards)
-            }
-            if self._evaluated_rewards  # store all the evaluated rewards
-            else {}
-        )
+        # # config_signature: tuple -> result: Dict
```
qingyun-wu, line 314: Remove?

qingyun-wu (approved):

PR #563, opened 2022-05-28
use zeroshot when no budget is given; custom_hp

sonichi:
* When no search budget is specified, train one model per estimator using zero-shot config and issue a warning (instead of setting time budget to 60s).
* Documentation updated.
* Test for no search budget and custom hp added.


qingyun-wu (approved):

liususan091219 (commented):
N/A

PR #567, opened 2022-06-02
version update

sonichi:
Update protobuf version to prevent TypeError in ray.


sonichi:
@Yard1 thanks for the prompt review!

Yard1:
FYI this requirement will be set in Ray in the next release.

PR #575, opened 2022-06-07
init value type match

sonichi:
1. allow init_value to be int when the domain is float.
2. change quniform in arima and sarimax to qrandint.


int-chaos (approved):

test/automl/test_regression.py
```diff
@@ -86,7 +87,11 @@ def test_regression(self):
         )
         print(automl.model.estimator)
         y_pred2 = automl.predict(X_train)
-        assert (y_pred == y_pred2).all()
+        # In some rare case, the last config is early stopped and it's the best config. But the logged config's ITER_HP is not reduced.
```
liususan091219, line 90: Is it better to replace ITER_HP with n_estimator, because the ITER_HP in Transformers is not n_estimator

sonichi: replaced.

qingyun-wu (approved):

liususan091219 (approved):

PR #584, opened 2022-06-12
backup & recover global vars for nested tune.run

### LINKED ISSUE #580: Nested FLAML optimizations get their search spaces tangled up
When using FLAML to search over models, which themselves have a FLAML regressor as a component, in the same Python process, we find that the two search spaces are confused, and the component FLAML regressors get fed arguments from the outer search space (and therefore fail).

We use joblib to work around that, so each inner FLAML regressor fit runs in a separate process spawned by joblib.

A way to reproduce is to run https://github.com/transferwise/auto-causality/blob/main/notebooks/Example.ipynb after setting `n_jobs=1` in [this line](https://github.com/transferwise/auto-causality/blob/d934d34c9ee9cba43f4ae9533836abc6dcb9ebf9/auto_causality/optimiser.py#L345) so the call runs in-process.

sonichi:
to resolve #580

### DIFF
--- a/flaml/tune/tune.py
+++ b/flaml/tune/tune.py
@@ -23,8 +23,6 @@ from .result import DEFAULT_METRIC
 import logging
 
 logger = logging.getLogger(__name__)
-
-
 _use_ray = True
 _runner = None
 _verbose = 0
@@ -299,6 +297,12 @@ def run(
     """
     global _use_ray
     global _verbose
+    global _running_trial
+    global _training_iteration
+    old_use_ray = _use_ray
+    old_verbose = _verbose
+    old_running_trial = _running_trial
+    old_training_iteration = _training_iteration
     if not use_ray:
         _verbose = verbose
         if verbose > 0:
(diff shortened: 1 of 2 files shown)


EgorKraevTransferwise (commented):
This will only work for one-level-deep nesting. If we're touching this at all, would it not be more elegant to use a stack structure, pushing the vars into it on start and popping them at the end, so we can do arbitrary nesting? :)

EgorKraevTransferwise:
I would rather do a try-catch, and in the catch block restore the global variables, then re-raise the exception. Then the original behavior is preserved exactly.

qingyun-wu (approved):

sonichi:
@EgorKraevTransferwise I think the PR is ready for merging. Would you like to test your use case by installing flaml from git and the branch nest before merging? Or would you like to test it after merging?

sonichi:
> I'd rather test after merging if that's all right

Merged. Please reopen #580 if it doesn't solve your problem.

EgorKraevTransferwise:
Tested it in auto-causality, runs without issues, thanks!

PR #591, opened 2022-06-15
Add python 3.10 in the CI

### LINKED ISSUE #308: Support Python 3.10
Python 3.10 has been released. We should test it. If all the dependencies support it, we should add it to CI.

sonichi:
close #308

### DIFF
--- a/test/nlp/test_default.py
+++ b/test/nlp/test_default.py
@@ -1,6 +1,5 @@
 from utils import get_toy_data_seqclassification, get_automl_settings
 import sys
-import pytest
 from flaml.default import portfolio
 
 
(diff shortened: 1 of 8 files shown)


test/nlp/test_default.py
```diff
@@ -16,6 +17,10 @@ def test_build_portfolio(path="./test/nlp/default", strategy="greedy"):
     portfolio.main()
 
 
+@pytest.mark.skipif(
```
qingyun-wu, line 20: Any reason why these tests should be disabled on mac os python 3.10?

qingyun-wu (approved):

liususan091219 (approved):

PR #596, opened 2022-06-16
trying to fix the indexerror for ner

liususan091219:
trying to fix the indexerror for ner

### DIFF
--- a/flaml/ml.py
+++ b/flaml/ml.py
@@ -176,13 +176,14 @@ def metric_loss_score(
                     ].mid.fmeasure
                 elif metric_name.startswith("seqeval"):
 
+                    label_len = len(labels)
                     zip_pred_true = [
                         [(p, lb) for (p, lb) in zip(prediction, label) if lb != -100]
                         for (prediction, label) in zip(y_predict, y_true)
                     ]
                     y_pred = [
                         [
-                            labels[p] if p in range(len(labels)) else -1
+                            labels[p] if (p < label_len and p > 0) else -1
                             for (p, l) in each_list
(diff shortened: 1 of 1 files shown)


flaml/ml.py
```diff
@@ -181,7 +181,10 @@ def metric_loss_score(
                         for (prediction, label) in zip(y_predict, y_true)
                     ]
                     y_pred = [
-                        [labels[p] for (p, l) in each_list]
+                        [
+                            labels[p] if p in range(len(labels)) else -1
```
sonichi, line 185: Please use a comparison expression instead of "in" which is much more efficient.
Also, remember to compute `len(labels)` out of the loop to be efficient.

sonichi (approved):
minor: 0 < p < label_len

flaml/ml.py
```diff
@@ -176,12 +176,13 @@ def metric_loss_score(
...
                 elif metric_name.startswith("seqeval"):
 
+                    label_len = len(labels)
                     zip_pred_true = [
                         [(p, lb) for (p, lb) in zip(prediction, label) if lb != -100]
                         for (prediction, label) in zip(y_predict, y_true)
                     ]
                     y_pred = [
-                        [labels[p] for (p, l) in each_list]
+                        [labels[p] if 0 < p < label_len else -1 for (p, l) in each_list]
```
sonichi, line 185: why not 0 <= p? does labels[0] correspond to -1?

sonichi (changes requested):

sonichi (approved):

PR #602, opened 2022-06-22
Fix automl settings in scikit-learn pipeline integration example

### LINKED ISSUE #601: Running tests in devcontainer creates unignored files
When working in a devcontainer, running tests creates a `core` file and `test` directory which are not included in the `.gitignore` file. Would be great to add them.

ZviBaratz:
Example should use a `settings` variable as suggested.


.gitignore
```diff
@@ -60,6 +60,8 @@ coverage.xml
 .hypothesis/
 .pytest_cache/
 cover/
+test/
```
sonichi, line 63: test/ is part of the repo.

sonichi (changes requested):
Looks good. One small change requested. Also, could you check if the top part of the notebook needs update?

.gitignore
ZviBaratz: Oops, I apologise about that - seems like I had a partial execution of the tests that never got to cleanup and therefore created a large number of files within my test/ directory.

qingyun-wu (approved):

PR #616, opened 2022-06-29
Allow custom GroupKFold object as split_type

### LINKED ISSUE #515: AttributeError: 'GroupKFold' object has no attribute 'groups'
When I set "split_type" as GroupKFold, then use fit() with setting "groups", I got the issue: _AttributeError: 'GroupKFold' object has no attribute 'groups'_ and

> ml.py:460, in evaluate_model_CV(config, estimator, X_train_all, y_train_all, budget, kf, task, eval_metric, best_val_loss, log_training_metric, fit_kwargs)
>     458     kf = kf.split(X_train_split, y_train_split)
>     459 elif isinstance(kf, GroupKFold):
> **_--> 460     groups = kf.groups_**
>     461     kf = kf.split(X_train_split, y_train_split, groups)
>     462     shuffle = False

So it seems the _groups_ value needs to be assigned in the **_evaluate_model_CV()_** function. Am I right?

sonichi:
close #515

Added additional checks for eval_method and split_type. When custom split_type is used, eval_method can't be "holdout".

### DIFF
--- a/test/automl/test_split.py
+++ b/test/automl/test_split.py
@@ -163,7 +163,6 @@ def test_object():
     automl = AutoML()
     automl_settings = {
         "time_budget": 2,
-        # "metric": 'accuracy',
         "task": "classification",
         "log_file_name": "test/{}.log".format(dataset),
         "model_history": True,
@@ -171,6 +170,9 @@ def test_object():
         "split_type": TestKFold(5),
     }
     automl.fit(X, y, **automl_settings)
+    assert (
+        automl._state.eval_method == "cv"
+    ), "eval_method must be 'cv' for custom data splitter"
 
 
 if __name__ == "__main__":
--- a/test/automl/test_split.py
+++ b/test/automl/test_split.py
@@ -1,6 +1,6 @@
 from sklearn.datasets import fetch_openml
 from flaml.automl import AutoML
-from sklearn.model_selection import train_test_split, KFold
+from sklearn.model_selection import GroupKFold, train_test_split, KFold
 from sklearn.metrics import accuracy_score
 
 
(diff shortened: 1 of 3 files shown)


wuchihsu:
Is it needed to also add eval_method test in `test_object()`?

sonichi:
> Is it needed to also add eval_method test in `test_object()`?

What about adding an assertion of `automl._state.eval_method=="cv"` after `automl.fit()`?

wuchihsu (approved):

test/automl/test_split.py
```diff
@@ -80,6 +80,19 @@ def test_groups():
     automl_settings["eval_method"] = "holdout"
```
qingyun-wu, line 80: Can eval_method be set to "auto"?

test/automl/test_split.py
```diff
@@ -80,6 +80,19 @@ def test_groups():
...
     automl.fit(X, y, **automl_settings)
 
+    automl_settings["split_type"] = GroupKFold(n_splits=3)
+    try:
+        automl.fit(X, y, **automl_settings)
+        raise RuntimeError(
+            "GroupKFold object as split_type should fail when eval_method is holdout"
+        )
+    except AssertionError:
```
qingyun-wu, line 89: Why catching AssertationError?

test/automl/test_split.py
sonichi: yes.

test/automl/test_split.py
sonichi: This is test code. We expect AssertionError to be raised in this case.

qingyun-wu (approved):

PR #627, opened 2022-07-05
isinstance(x, int) -> isinstance(x, (int, np.integer))

liususan091219:
isinstance(x, int) -> isinstance(x, (int, np.integer))


test/automl/test_notebook_example.py
```diff
@@ -148,4 +148,4 @@ def test_mlflow():
 
 
 if __name__ == "__main__":
-    test_automl(600)
+    test_automl(5)
```
sonichi, line 151: don't modify this line

sonichi (approved):
Please create a new branch from main when you create a new PR.

PR #628, opened 2022-07-07
Disable max_seq_length in NER to temporarily prevent the inconsistency issue. 

liususan091219:
Disable max_seq_length in NER to temporarily prevent the inconsistency issue.


sonichi:
Please create a new branch from main when you create a PR.

PR #629, opened 2022-07-07
disable max_len for ner

liususan091219:
disable max_len for ner


sonichi (commented):
This will make max_seq_length in hf_args invalid, right? Need to update the corresponding parts of code and documentation.

liususan091219:
> This will make max_seq_length in hf_args invalid, right? Need to update the corresponding parts of code and documentation.

No, it only disables NER, since NER is where the inconsistency problem happens (i.e., the ground truth is changed).

sonichi:
> > This will make max_seq_length in hf_args invalid, right? Need to update the corresponding parts of code and documentation.
> 
> No, it only disables NER, since NER is where the inconsistency problem happens (i.e., the ground truth is changed).

I mean this behavior is not documented and users don't know it's ignored even when they set this argument.

flaml/nlp/huggingface/training_args.py
```diff
@@ -49,7 +49,7 @@ class TrainingArgumentsForAuto(TrainingArguments):
 
     fp16: bool = field(default=True, metadata={"help": "whether to use the FP16 mode"})
 
-    max_seq_length: int = field(default=128, metadata={"help": "max seq length"})
+    max_seq_length: int = field(default=None, metadata={"help": "max seq length"})
```
sonichi, line 52: Would other tasks be affected by this change?

flaml/nlp/utils.py
```diff
@@ -12,6 +12,10 @@
 )
 
 import pandas as pd
+import logging
+
+logger = logging.getLogger("flaml.automl")
+FREE_MEM_RATIO = 0.2
```
sonichi, line 18: What for?

flaml/nlp/utils.py
```diff
@@ -12,6 +12,10 @@
 )
 
 import pandas as pd
+import logging
+
+logger = logging.getLogger("flaml.automl")
```
sonichi, line 17: why "flaml.automl"?

flaml/nlp/utils.py
```diff
@@ -85,13 +89,19 @@ def tokenize_and_align_labels(
     Y_sent_key=None,
     return_column_name=False,
 ):
+    if hf_args.max_seq_length is not None:
+        logger.warning(
```
sonichi, line 93: shall this warning be issued only if task is token classification?

flaml/nlp/utils.py
```diff
@@ -85,13 +89,19 @@ def tokenize_and_align_labels(
...
+            )
+        )
     tokenized_inputs = tokenizer(
         [list(examples[X_sent_key])],
         padding="max_length"
         if hf_args and hf_args.pad_to_max_length
         else False,  # to be consistent with https://github.com/huggingface/transformers/blob/main/examples/pytorch/token-classification/run_ner.py#L394
         truncation=True,
-        max_length=hf_args.max_seq_length if hf_args else None,
+        max_length=None,  # hf_args.max_seq_length if hf_args else None,
```
sonichi, line 104: This is always set to None, not only for token classification task. This contradicts with the documentation in training_args.py.

sonichi (commented):
The documentation contradicts with the code. The max_length parameter seems to be always ignored without checking whether the task is token classification.
If I misunderstand, please clarify and add the clarification comments in the code.

flaml/nlp/utils.py
liususan091219: I think it's within the tokenize_and_align_labels function so only for token classification task.

flaml/nlp/utils.py
liususan091219: I think it's within the tokenize_and_align_labels function so only for token classification task.

flaml/nlp/utils.py
liususan091219: I just saw this line from model.py so just trying to copy it. If it's not flaml.automl, what should it be?

flaml/nlp/huggingface/training_args.py
liususan091219: For other tasks, the default max_seq_length will now be None; however, the user can still set max_seq_length for the other tasks.

sonichi (commented):
Please add comments in the code as suggested below.

flaml/nlp/utils.py
```diff
@@ -85,13 +88,16 @@ def tokenize_and_align_labels(
     Y_sent_key=None,
     return_column_name=False,
 ):
+    logger.warning(
+        "For token classification task, FLAML currently does not support customizing the max_seq_length, so max_seq_length will always be None."
+    )
```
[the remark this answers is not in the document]
liususan091219: Now the default max_seq_length is not None (because when it is None, the pytest will fail), so the if condition is always true.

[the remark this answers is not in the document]
sonichi: "Also, I think we only need to issue it once for the task. No need to issue it every time this function is called."

flaml/nlp/huggingface/utils.py
```diff
@@ -0,0 +1,464 @@
...
+    return_column_name=False,
+):
+    # tokenize_and_align_labels is only called by the token-classification task
+    tokenized_inputs = tokenizer(
+        [list(examples[X_sent_key])],
+        padding="max_length"
+        if hf_args and hf_args.pad_to_max_length
+        else False,  # to be consistent with https://github.com/huggingface/transformers/blob/main/examples/pytorch/token-classification/run_ner.py#L394
+        truncation=True,
+        max_length=hf_args.max_seq_length,  # hf_args.max_seq_length if hf_args else None,
```
sonichi, line 79:
```suggestion
        max_length=hf_args.max_seq_length if hf_args else None,
```

sonichi (approved):

PR #645, opened 2022-07-21
Fix HPO evaluation bug

### LINKED ISSUE #637: AutoML's best val_loss is still incorrect for TransformersEstimator
There's a bug that makes AutoML's best val_loss is still incorrect (and larger) for TransformersEstimator.

liususan091219:
1. Update the search space to HF's
2. Update eval steps to 500
3. Update HPO selection criterion to the HPO metric to fix the bug in issue (Closes #637)
4. Update automl_nlp.ipynb for KDD tutorial.


flaml/model.py
```diff
@@ -514,7 +500,7 @@ def _set_training_args(self, **kwargs):
 
         self._training_args.eval_steps = (
             self._training_args.logging_steps
-        ) = self._training_args.saving_steps = self.checkpoint_freq
+        ) = self._training_args.saving_steps = 500  # self.checkpoint_freq
```
sonichi, line 503: why fixing it as a constant?

sonichi (commented):
The notebook is too long to review. Maybe remove the very long output for some cells. Then I can review it.

flaml/model.py
liususan091219: So the evaluation frequency is not too high.

sonichi: my question is why fixing it a constant.
The original code uses `self._checkpoint_freq`. If you keep using `self._checkpoint_freq`, can't you make "evaluation frequency not too high"?

liususan091219: I want to propose to remove self._checkpoint_freq, it’s less interpretable than eval_steps setting as a constant.

sonichi: Then remove it and write it down. Right now it's not removed and your proposal is not written down in the PR.

flaml/model.py
```diff
@@ -417,32 +417,18 @@ def _join(X_train, y_train, task):
...
-            "weight_decay": {
-                "domain": tune.choice([0, 0.1, 0.2, 0.3]),
-                "init_value": 0.0,
-            },
-            "adam_epsilon": {
-                "domain": tune.choice([1e-8, 1e-7, 1e-6]),
-                "init_value": 1e-6,
-            },
-            "seed": {"domain": tune.randint(40, 45), "init_value": 42},
+            "seed": {"domain": tune.randint(1, 40), "init_value": 20},
```
sonichi, line 431: I am not sure about replacing the old search space completely. Shall we keep the old search space in an alternative estimator? We'll need to compare the two search spaces later.

test/nlp/test_default.py
```diff
@@ -91,9 +91,14 @@ def test_points_to_evaluate():
...
     automl_settings["estimator_list"] = ["transformer_ms"]
-    automl_settings["starting_points"] = "data"
+    automl_settings["starting_points"] = "data:test/nlp/default/"
 
     del automl_settings["fit_kwargs_by_estimator"]["transformer_ms"]["model_path"]
+    automl_settings["custom_hp"] = {
+        "transformer_ms": {
+            "model_path": {"domain": "google/electra-small-discriminator"}
+        }
+    }
```
sonichi, line 101: these are inconsistent with #647

sonichi (commented):
The output of some cells is still too long. Could you remove the long output?

flaml/model.py
liususan091219: We should do that in the experiment branch and finalize with the one that wins. Here for the stable version it's better to fix to one space so users are not confused.

test/nlp/test_default.py
liususan091219: > The output of some cells is still too long. Could you remove the long output?

I already set transformers.logging.set_verbosity_error(), shall I manually remove?

liususan091219: > these are inconsistent with #647

reverted

flaml/model.py
sonichi: Have you confirmed the new search space wins? I thought it's not settled yet.

liususan091219: It is better than the discrete space on mrpc.

sonichi (commented):
I suggest keeping at least one simple example without ray so that they can test it quickly. And explain the difference between using and not using ray. For the example which requires long time to finish, give users a headsup.

sonichi (approved):

PR #647, opened 2022-07-23
This PR fixes the frequent NLP bugs in the other PRs

liususan091219:
Disable py3.6 for text summarization to fix the NLP bug in the other PRs

### DIFF
--- a/test/nlp/test_default.py
+++ b/test/nlp/test_default.py
@@ -114,11 +114,6 @@ def test_zero_shot_nomodel():
     automl_settings = get_automl_settings(estimator_name)
 
     del automl_settings["fit_kwargs_by_estimator"][estimator_name]["model_path"]
-    automl_settings["custom_hp"] = {
-        "transformer_ms": {
-            "model_path": {"domain": "google/electra-small-discriminator"}
-        }
-    }
 
     (
         hyperparams,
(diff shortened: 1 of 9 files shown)


test/nlp/test_default.py
```diff
@@ -91,9 +91,13 @@ def test_points_to_evaluate():
     automl_settings = get_automl_settings(estimator_name="transformer_ms")
 
     automl_settings["estimator_list"] = ["transformer_ms"]
-    automl_settings["starting_points"] = "data"
-
+    automl_settings["starting_points"] = "data:test/nlp/default/"
     del automl_settings["fit_kwargs_by_estimator"]["transformer_ms"]["model_path"]
```
sonichi, line 95: Why did we have "model_path" in `fit_kwargs_by_estimator` in the beginning? Shall we remove it in `get_automl_settings()`?

liususan091219: get automl settings is for providing a low cost config so it includes overwriting the model to Electra small

sonichi: I said `fit_kwargs_by_estimator`. Why does it contain "model_path"?

liususan091219: We must overwrite model_path to electra small. Do you mean to overwrite "model_path", why do we do it via fit_kwargs_by_estimator instead of other options (such as fit_kwargs)?

sonichi (approved):

Issue #660, opened 2022-08-03
Disable data validation


jmrichardson:
I would like to be able to disable pre-processing of data as I am working with a very large dataset.  The dataset is clean and does not contain missing values.   However, it appears that flaml iterates each feature (1000s) with millions of observations to replace missing data:

`X[column] = X[column].fillna(np.nan)`

In addition it does simple imputation:

```
self.transformer = ColumnTransformer(
                    [
                        (
                            "continuous",
                            SimpleImputer(missing_values=np.nan, strategy="median"),
                            X_num.columns,
                        )
                    ]
                )
                X[num_columns] = self.transformer.fit_transform(X_num)
```

This takes a considerable amount of time and would prefer to disable data preprocessing.  Thank you for your consideration.

sonichi:
The solution would be:
1. https://github.com/microsoft/FLAML/blob/21fa6c10ec6f963edbda99e9020ad585ae60a1a5/flaml/automl.py#L1122 -> `if issparse(X_train_all) or self._skip_transform:`
2. add a keyword argument in both `AutoML.__init__()` and `AutoML.fit()` in automl.py and set `self._skip_transform` to it before https://github.com/microsoft/FLAML/blob/21fa6c10ec6f963edbda99e9020ad585ae60a1a5/flaml/automl.py#L2429 You can refer to how other keyword arguments are handled in these two places.

jmrichardson:
Hi @sonichi 

Should I create a PR for this change?  Happy to give a shot tomorrow.

sonichi:
> Hi @sonichi
> 
> Should I create a PR for this change? Happy to give a shot tomorrow.

Yes please. :)

jmrichardson:
Hi, I am not an expert in PRs and not sure if the first PR should be applied first?  If I make the modifications for this and submit a new PR I will have the updates from the first PR included in the second PR.  Is this acceptable?

sonichi:
Thanks for the question. As long as the second PR doesn't depend on the first PR, you don't need to wait for the first PR to be merged. You can create the second PR still from the main branch and it shouldn't include the updates from the first PR. When either PR is merged to main, you can update the other PR by merging from the (new) main into the PR branch. Is that clear?

PR #665, opened 2022-08-04
Skip transform

jmrichardson:
#660


flaml/automl.py
```diff
@@ -665,6 +665,7 @@ def custom_metric(
                 provided in custom_hp. Note that the value for "domain" can either be a constant
                 or a sample.Domain object.
                 e.g.,
+            skip_transform: boolean, default=False | Whether to pre-process data prior to modeling
```
sonichi, line 668: Please move this line after 681. Line 669-081 is part of the example for `custom_hp`.
Also, please add a period in the end.

flaml/automl.py
```diff
@@ -2273,6 +2276,9 @@ def custom_metric(
             custom_hp: dict, default=None | The custom search space specified by user
                 Each key is the estimator name, each value is a dict of the custom search space for that estimator. Notice the
                 domain of the custom search space can either be a value of a sample.Domain object.
+            skip_transform: boolean, default=False | Whether to pre-process data prior to modeling
+
+
```
sonichi, line 2281: Similar to the previous comment.

flaml/automl.py
```diff
@@ -2418,6 +2424,7 @@ def custom_metric(
 
         self._state.fit_kwargs = fit_kwargs
         custom_hp = custom_hp or self._settings.get("custom_hp")
+        self._skip_transform = skip_transform or self._settings.get("skip_transform")
```
sonichi, line 2427: Since `skip_transform` is a boolean, please follow line 2387's logic to check whether it's None instead of following line 2426. Otherwise, when skip_transform is False, its value will be overridden by self._settings.

sonichi (commented):
Please add a test for the case of skipping transform. For example, add a `test_nopreprocess` test in https://github.com/microsoft/FLAML/blob/21fa6c10ec6f963edbda99e9020ad585ae60a1a5/test/automl/test_classification.py#L30

flaml/automl.py
jmrichardson: whoops, sorry about that

flaml/automl.py
jmrichardson: Ok updated the logic and also added test

flaml/automl.py
```diff
@@ -678,6 +678,7 @@ def custom_metric(
              }
          }
         ```
+            skip_transform: boolean, default=False | Whether to pre-process data prior to modeling
```
sonichi, line 681:
```suggestion
            skip_transform: boolean, default=False | Whether to pre-process data prior to modeling.
```

flaml/automl.py
```diff
@@ -2287,6 +2292,7 @@ def custom_metric(
         }
         ```
 
+        skip_transform: boolean, default=False | Whether to pre-process data prior to modeling
```
sonichi, line 2295:
```suggestion
        skip_transform: boolean, default=False | Whether to pre-process data prior to modeling.
```

sonichi (commented):
Looks good except a minor thing.

flaml/automl.py
jmrichardson: Sounds good

sonichi (commented):
`self._skip_transformer` is also needed before line 1657, hence the argument needs to be added to `retrain_from_log()`.

flaml/automl.py
```diff
@@ -2418,6 +2424,7 @@ def custom_metric(
 
         self._state.fit_kwargs = fit_kwargs
         custom_hp = custom_hp or self._settings.get("custom_hp")
+        self._skip_transform = self._settings.get("skip_transform") if skip_transform is None else skip_transform
```
liususan091219, line 2427:
```suggestion
        self._skip_transform = skip_transform or self._settings.get("skip_transform")
```

liususan091219 (changes requested):

flaml/automl.py
sonichi: @liususan091219 See my comment in https://github.com/microsoft/FLAML/pull/665#discussion_r938235656

liususan091219: resolved.

liususan091219 (commented):
My comment was resolved. 

Kindly be reminded there are still unresolved comments in this PR: https://github.com/microsoft/FLAML/pull/665#pullrequestreview-1062710393, please take care of them when you have time.

jmrichardson:
> `self._skip_transformer` is also needed before line 1657, hence the argument needs to be added to `retrain_from_log()`. That's the reason for the failure in https://github.com/microsoft/FLAML/runs/7700225039?check_suite_focus=true#step:9:1951

Updated and hopefully fixes the issue

test/automl/test_classification.py
```diff
@@ -158,6 +158,23 @@ def test_preprocess(self):
         automl.fit(X, y, **automl_settings)
```
liususan091219, line 158:
```suggestion
        automl_settings["keep_search_state"] = True
        automl.fit(X, y, **automl_settings)
        X, y = automl._X_train_all, automl._y_train_all
```

liususan091219 (changes requested):

test/automl/test_classification.py
liususan091219: You need to pass a transformed data before you can skip the data transform.

liususan091219 (approved):

PR #669, opened 2022-08-07
Support customized cross-validation strategy

skzhang1:


flaml/automl.py
```diff
@@ -2070,6 +2071,7 @@ def fit(
         use_ray=None,
         metric_constraints=None,
         custom_hp=None,
+        cv_strategy=None,
```
sonichi, line 2074: what about "cv_score_agg_func"?
@jmrichardson

flaml/automl.py
```diff
@@ -2287,6 +2289,17 @@ def custom_metric(
         }
         ```
 
+        cv_strategy: customized function, the strategy of conducting cross-validation. Default to average the optimization metric across folds.
+        We give an example here:
+
+        ```python
+        def cv_strategy(val_loss_folds):
+            return sum(val_loss_folds)/len(val_loss_folds)
```
sonichi, line 2297: Please clarify the input and output.
My understanding is that the input is a list of 2-tuples. Each tuple has two elements, the first being a loss score to minimize and the second being a dict of all the metrics to log or None.
And the output is a 2-tuple with the same assumption similar to the "metric" function.

Please refer to the docstr of "metric".
Also, please add this argument in the docstr and the implementation of the constructor of `AutoML` similar to other kwargs.

flaml/automl.py
```diff
@@ -2447,6 +2460,7 @@ def custom_metric(
         eval_method = self._decide_eval_method(eval_method, time_budget)
         self._state.eval_method = eval_method
         logger.info("Evaluation method: {}".format(eval_method))
+        self._state.cv_strategy = cv_strategy
```
sonichi, line 2463: Change this line after you change the constructor of `AutoML`. Refer to line 2433 for example.

sonichi (commented):
Please make sure you get approvals from @jmrichardson and @qingyun-wu

flaml/automl.py
skzhang1: _"the second being a dict of all the metrics to log or None"_, What does this mean? Do we need to support customizing cv_score_agg_func for other metrics? I think in most cases, users only want to customize CV for one metric

sonichi: We allow users to define one metric to optimize for and multiple metrics to log in custom "metric" function (https://microsoft.github.io/FLAML/docs/Use-Cases/Task-Oriented-AutoML#optimization-metric). So I wonder that we should allow the same in the custom "cv_score_agg_func".

flaml/ml.py
```diff
@@ -438,11 +438,28 @@ def evaluate_model_CV(
...
+                if single_fold[1] is None:
+                    break
+                elif metrics_to_log is None:
+                    metrics_to_log = single_fold[1]
+                else:
+                    metrics_to_log = {k: metrics_to_log[k] + v for k, v in single_fold[1].items()}
+            if metrics_to_log:
+                n = len(metrics_across_folds)
+                metrics_to_log = {k: v / n for k, v in metrics_to_log.items()}
+            return metric_to_minimize, metrics_to_log
```
sonichi, line 459: Defining here will slow it down every time the function `evaluate_model_CV()` is called. Move the definition out.

flaml/ml.py
```diff
@@ -515,27 +532,28 @@ def evaluate_model_CV(
...
             elif total_metric is not None:
                 total_metric += metric_i
             else:
                 total_metric = metric_i
         train_time += train_time_i
         pred_time += pred_time_i
         if valid_fold_num == n:
-            val_loss_list.append(total_val_loss / valid_fold_num)
-            total_val_loss = valid_fold_num = 0
+            val_loss_list.append(cv_score_agg_func(list(zip(val_loss_folds,[None]*len(val_loss_folds))))[0])
```
sonichi, line 546: Aggregate after all the folds finished. Call the aggregation function only once.

flaml/ml.py
```diff
@@ -515,27 +532,28 @@ def evaluate_model_CV(
...
+        val_loss_folds.append(val_loss_i)
         if log_training_metric or not isinstance(eval_metric, str):
-            if isinstance(total_metric, dict):
-                total_metric = {k: total_metric[k] + v for k, v in metric_i.items()}
+            if isinstance(metric_i, dict):
+                log_metric_folds.append(metric_i)
             elif total_metric is not None:
                 total_metric += metric_i
             else:
                 total_metric = metric_i
```
sonichi, line 542: remove?

flaml/ml.py
```diff
@@ -515,27 +532,28 @@ def evaluate_model_CV(
...
             break
     val_loss = np.max(val_loss_list)
     n = total_fold_num
     if log_training_metric or not isinstance(eval_metric, str):
-        if isinstance(total_metric, dict):
-            metric = {k: v / n for k, v in total_metric.items()}
+        if len(log_metric_folds):
+            metric = cv_score_agg_func(list(zip([0]*len(log_metric_folds),log_metric_folds)))[1]
         else:
             metric = total_metric / n
```
sonichi, line 558: These need to be updated after you address the previous comment.

flaml/ml.py
```diff
@@ -438,11 +438,28 @@ def evaluate_model_CV(
     task,
     eval_metric,
     best_val_loss,
+    cv_score_agg_func,
```
liususan091219, line 441:
```suggestion
    cv_score_agg_func=default_cv_score_agg_func,

```

flaml/ml.py
liususan091219: Maybe move out cv_score_agg_func (before evaluate_model_CV) and rename it default_cv_score_agg_func?

liususan091219 (changes requested):

flaml/ml.py
skzhang1: Do we allow metric_i to be a numeric value?

flaml/ml.py
```diff
@@ -516,31 +532,20 @@ def evaluate_model_CV(
...
+        val_loss_folds.append(val_loss_i)
         if log_training_metric or not isinstance(eval_metric, str):
-            if isinstance(total_metric, dict):
-                total_metric = {k: total_metric[k] + v for k, v in metric_i.items()}
-            elif total_metric is not None:
-                total_metric += metric_i
-            else:
-                total_metric = metric_i
+            if isinstance(metric_i, dict):
+                log_metric_folds.append(metric_i)
```
sonichi, line 539: 1. remove the "if" condition since we need to append metric_i to the list anyway.
2. metric_i can be a single float number as well. So the list of `log_metric_folds` can contain dicts or floats or Nones.

sonichi (changes requested):

flaml/ml.py
sonichi: yes.

flaml/ml.py
```diff
@@ -516,31 +532,20 @@ def evaluate_model_CV(
...
-    val_loss = np.max(val_loss_list)
-    n = total_fold_num
     if log_training_metric or not isinstance(eval_metric, str):
-        if isinstance(total_metric, dict):
-            metric = {k: v / n for k, v in total_metric.items()}
-        else:
-            metric = total_metric / n
+        val_loss, metric = cv_score_agg_func(list(zip([0]*len(log_metric_folds),log_metric_folds)))
+    else:
+        val_loss, metric = cv_score_agg_func(list(zip(val_loss_folds,[None]*len(val_loss_folds))))
```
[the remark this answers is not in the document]
skzhang1: Yes

flaml/automl.py
```diff
@@ -2350,6 +2357,38 @@ def custom_metric(
...
         ```
 
+            cv_score_agg_func: customized cross-validation scores aggregate function. Default to average metrics across folds. If specificed, this function needs to
+                have the following signature:
+
+        ```python
+        def cv_score_agg_func(val_loss_folds, log_metrics_folds):
+            return metric_to_minimize, metrics_to_log
+        ```
+                “val_loss_folds” - list of float, it records the loss scores of each ford; “log_metrics_folds” - list of dict/float, it records the metrics of each fords to log.
```
sonichi, line 2367:
```suggestion
                “val_loss_folds” - list of floats, the loss scores of each fold; “log_metrics_folds” - list of dicts/floats, the metrics of each fold to log.
```

flaml/automl.py
```diff
@@ -2350,6 +2357,38 @@ def custom_metric(
...
+            for single_fold in log_metrics_folds:
+                if metrics_to_log is None:
+                    metrics_to_log = single_fold
+                elif isinstance(metrics_to_log, dict):
+                    metrics_to_log = {k: metrics_to_log[k] + v for k, v in single_fold.items()}
+                else:
+                    metrics_to_log += single_fold
+            if metrics_to_log:
+                n = len(val_loss_folds)
+                metrics_to_log = {k: v / n for k, v in metrics_to_log.items()} if isinstance(metrics_to_log, dict) else metrics_to_log/n
```
sonichi, line 2384:
```suggestion
                metrics_to_log = {k: v / n for k, v in metrics_to_log.items()} if isinstance(metrics_to_log, dict) else metrics_to_log / n
```

flaml/ml.py
sonichi: This comment hasn't been addressed.

sonichi (commented):
Previous comments haven't been addressed.

skzhang1:
Has addressed @sonichi

flaml/ml.py
```diff
@@ -541,8 +541,7 @@ def evaluate_model_CV(
         total_fold_num += 1
         val_loss_folds.append(val_loss_i)
         if log_training_metric or not isinstance(eval_metric, str):
-            if isinstance(metric_i, dict):
-                log_metric_folds.append(metric_i)
+            log_metric_folds.append(metric_i)
```
sonichi, line 544: I think this "if" needs to be removed too.

sonichi (approved):

flaml/ml.py
skzhang1: I think we can not remove this "if".  @sonichi 
1. If users do not want log_training_metric (set it to False), this "if" could save some computation resources.
2. sometimes, metric dicts are not norm. It can be {"metric1": float, "metric2": list}. [test/automl](https://github.com/microsoft/FLAML/blob/main/test/automl/test_custom_hp.py#:~:text=def%20test_custom_hp_nlp()%3A  ). It will be hard for us to set default_cv_score_agg_func

qingyun-wu (approved):

flaml/ml.py
sonichi: 1. Even when log_training_metric is False, `metric_i` is not necessarily None, e.g., containing "pred_time".
2. The case of `"metric2": list` is problematic. @liususan091219 could you propose a solution?

skzhang1: @liususan091219 Thanks! It fixed now.

PR #673, opened 2022-08-09
replaced set with ordered set to ensure reproduciblity

AdityaN1198:
Replaced python 'set' with OrderedSets from "ordered-set" library to ensure reproducibility during runs when seed is set.
Added ordered-set in setup.py as well


sonichi:
Could you describe the problem when not using OrderedSet? Do you have an example of using OrderedSet instead of set to make an irreproducible result reproducible? Thanks.

PR #677, opened 2022-08-10
categorical choice can be ordered or unordered

sonichi:
Add argument `order` to the `tune.choice()` function. Allow users to specify whether the choices have an order or not. This is useful for numerical choices like "seed". When not specified, the numerical choices will be regarded as having an order.


flaml/tune/space.py
```diff
@@ -225,15 +225,18 @@ def add_cost_to_space(space: Dict, low_cost_point: Dict, choice_cost: Dict):
...
-                domain.ordered = True
             else:
-                domain.ordered = False
+                ordered = getattr(domain, "ordered", None)
+                if ordered is None:
+                    # automatically decide whether to order the choices based on the value type
+                    domain.ordered = ordered = all(
+                        isinstance(x, (int, float)) for x in domain.categories
+                    )
+                if ordered:
```
liususan091219, line 235:
```suggestion
                if ordered is True:

```

liususan091219: add is True to tell apart from line 230

liususan091219 (changes requested):

flaml/tune/space.py
sonichi: What do you mean by "tell apart from line 230"?

liususan091219: line 230 is "if ordered is None". If what you mean here is "if ordered is True" instead of "if ordered is not None", could make it more clear.

qingyun-wu (approved):

flaml/tune/space.py
sonichi: `ordered` is bool or None. `if ordered` means `if ordered is True`. There is no ambiguity here.
In contrast, if we want to check whether `ordered` is False, we need to explicitly write `ordered is False` instead of `not ordered` because the latter indicates `ordered` is either False or None.
These two situations are different.

liususan091219 (approved):

PR #686, opened 2022-08-15
VW version requirement and documentation on config_constraints vs metric_constraints

qingyun-wu:
1. Adding VW version requirements to ensure AutoVW's compatibility with VW.
2. Add documentation on config_constraints and metric_constraints. 

## Checks


website/docs/Use-Cases/Tune-User-Defined-Function.md
```diff
@@ -265,24 +265,27 @@ A user can specify constraints on the configurations to be satisfied via the arg
...
 ```python
-flaml.tune.run(evaluation_function=evaluate_config, mode="min",
-               config=config_search_space,
-               metric_constraints=[("score", "<=", 0.4)],...)
+analysis = tune.run(...,
+    metric_constraints = [("training_cost", "<=", 1)]),
 ```
 
+#### **`config_constraints` vs `metric_constraints`:**
+The key difference between these two types of constraints is that the calculation of constraints in `config_constraints` **does not rely on the evaluation procedure in the metric function**. For example, when a constraint only depends on the config itself, as shown in the code example. Due to this independency, constraints in `config_constraints` will be checked before evaluation. So configurations that do not satisfy `config_constraints` will not be evaluated.
```
sonichi, line 286: What's "metric function"? Could you rephrase it to avoid confusion?

sonichi (commented):
Please make the auto checks pass.

olgavrou (approved):

sonichi (approved):

website/docs/Use-Cases/Tune-User-Defined-Function.md
```diff
@@ -265,24 +265,27 @@ A user can specify constraints on the configurations to be satisfied via the arg
...
 ```python
-flaml.tune.run(evaluation_function=evaluate_config, mode="min",
-               config=config_search_space,
-               metric_constraints=[("score", "<=", 0.4)],...)
+analysis = tune.run(...,
+    metric_constraints = [("training_cost", "<=", 1)]),
 ```
 
+#### **`config_constraints` vs `metric_constraints`:**
+The key difference between these two types of constraints is that the calculation of constraints in `config_constraints` does not rely on the computation procedure in the evaluation function, i.e., in `evaluation_function`. For example, when a constraint only depends on the config itself, as shown in the code example. Due to this independency, constraints in `config_constraints` will be checked before evaluation. So configurations that do not satisfy `config_constraints` will not be evaluated.
```
YIWEI-CHEN, line 286: My suggestion:
`metric_constraints` need model's output or statistic of evaluation function (e.g., training cost) and are computed in the `evaluation_function`. `config_constraints` only depend on the config itself, as shown in the model size example. `config_constraints` does not rely on the computation procedure in `evaluation_function`. Due to this independency, `config_constraints` will be checked before evaluation. The configurations that do not satisfy `config_constraints` will not be used in `evaluation_function`.

YIWEI-CHEN (commented):
My suggestion for `config_constraints` vs `metric_constraints`:
`metric_constraints` need model's output (e.g., Precision) or statistic of evaluation function (e.g., training cost). They are computed in the `evaluation_function`. `config_constraints` only depend on the config itself, as shown in the model size example. `config_constraints` does not rely on the computation procedure in evaluation_function, so config_constraints will be checked before evaluation. The configurations that do not satisfy config_constraints are not used in `evaluation_function`.

PR #719, opened 2022-09-07
Dockerfile building problem

skzhang1:
Dockerfile building problem.


Dockerfile
```diff
@@ -17,11 +17,11 @@ WORKDIR /home/flaml-dev/FLAML
 RUN sudo pip install -e .[test,notebook]
 
 # Install precommit hooks
-RUN pre-commit install
+RUN sudo pre-commit install
```
qingyun-wu, line 20: Is this sudo necessary?

skzhang1: resolve

sonichi (approved):

PR #747, opened 2022-10-03
log search_state.config directly instead of under tag config

prithvikannan:
Signed-off-by: Prithvi Kannan <prithvi.kannan@databricks.com>

Currently, FLAML logs a parameter called "config" with dictionary of all config parameters MLflow for each run, creating a "key-dictionary" pair. This makes further analysis using MLflow difficult, such as comparing runs, which is designed to compare key-value pairs. 

In this change, we propose logging the parameters directly to MLflow, rather than bundled into the "config" parameter. `best_config` will continue to be logged as a dict for consistency.

## Checks

### DIFF
--- a/flaml/automl.py
+++ b/flaml/automl.py
@@ -3108,7 +3108,7 @@ class AutoML(BaseEstimator):
                 mlflow.log_metric("trial_time", search_state.trial_time)
                 mlflow.log_metric("wall_clock_time", self._state.time_from_start)
                 mlflow.log_metric("validation_loss", search_state.val_loss)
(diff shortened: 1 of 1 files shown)


qingyun-wu (approved):

liususan091219 (approved):

liususan091219:
> @liususan091219 does this error look flaky to you? seems like the build on windows worked fine, but fails on macos. can you retrigger the test suite (I don't have permission to do so)

@prithvikannan no errors now.

@sonichi @qingyun-wu can you merge the PR?

PR #756, opened 2022-10-10
Remove NLP classification head

liususan091219:

When the target task has a different number of classes than the model loaded, there is a dimension mismatch problem. Previously this problem was handled by calling the ClassificationHead class, e.g., ElectraClassificationHead. This method is not general as we need to define a mapping to keep track of which model needs a classification head. 

This PR removes such mapping. Instead adding the argument `ignore_mismatched_sizes=True.` 

+ Adding tests for switching classification count from {1,2,3} -> {1,2,3}
+ Skipping Windows tests for the newly added test `test/nlp/test_autohf_classificationhead.py`
+ Update the return metric of `test_autohf_custom_metric.py` to `eval_runtime` to avoid the key not found error. 
+ Cleaning the output file

## Checks


sonichi (commented):
Could you add test for the head size mismatch situations? For example, 3->2, 2->3, 3->1, 1->3, 2->1.

test/nlp/test_autohf_classificationhead.py
```diff
@@ -1,26 +1,52 @@
...
+            try:
+                automl.fit(
+                    X_train=X_train,
+                    y_train=y_train,
+                    X_val=X_val,
+                    y_val=y_val,
+                    **automl_settings
+                )
+            except requests.exceptions.HTTPError:
+                return
```
sonichi, line 48:
```suggestion
                continue
```

flaml/model.py
```diff
@@ -667,6 +667,7 @@ def on_epoch_end(self, args, state, control, **callback_kwargs):
                     control.should_save = True
                     control.should_evaluate = True
 
+        self._training_args.report_to = []
```
sonichi, line 670: Please add comment why this line is needed.

test/nlp/test_autohf.py
```diff
@@ -71,6 +71,12 @@ def test_hf_data():
 
     del automl
 
+    import os
+    import shutil
+
+    if os.path.exists("test/data/output/"):
+        shutil.rmtree("test/data/output/")
```
sonichi, line 78: Is it a good idea to add in the documentation to tell users that the output may consume large space on disk and how to clean it etc.? Could create another issue.

test/nlp/test_autohf_classificationhead.py
```diff
@@ -1,14 +1,104 @@
...
+]
+model_path_list = [
+    "textattack/bert-base-uncased-STS-B",
+    "textattack/bert-base-uncased-SST-2",
+    "textattack/bert-base-uncased-MNLI",
+]
 
-def test_classification_head():
+
+@pytest.mark.skipif(sys.platform.startswith("win"), reason="do not run on windows")
```
sonichi, line 22: Is the intention to skip this single test on windows and execute all the other tests? Why?

test/nlp/test_autohf_custom_metric.py
```diff
@@ -36,7 +36,7 @@ def custom_metric(
     metrics = trainer.evaluate(eval_dataset)
     estimator._metric = estimator_metric_backup
 
-    return metrics.pop("eval_automl_metric"), metrics
+    return metrics.pop("eval_runtime"), metrics
```
sonichi, line 39: Could you explain the reason of this change?

test/nlp/test_default.py
```diff
@@ -79,6 +79,11 @@ def test_starting_point_not_in_search_space():
         automl._search_states[this_estimator_name].search_space["model_path"]
         == "albert-base-v2"
     )
+    import os
+    import shutil
```
sonichi, line 83: move to the beginning

test/nlp/test_autohf_tokenclassification.py
```diff
@@ -62,6 +62,12 @@ def test_tokenclassification_idlabel():
                 if min_inter_result != sys.maxsize:
                     assert val_loss == min_inter_result
 
+    import os
+    import shutil
```
sonichi, line 66: move to the beginning

test/nlp/test_autohf_classificationhead.py
liususan091219: Yes, this is the newly added test, and I am currently skipping it on windows because it fails for pytests on windows platforms.

test/nlp/test_autohf_custom_metric.py
liususan091219: Update the return metric of test_autohf_custom_metric.py to eval_runtime to avoid the key not found error. It is explained above in the description of the PR.

flaml/model.py
liususan091219: Removed

test/nlp/test_autohf.py
liususan091219: Added.

test/nlp/test_autohf_classificationhead.py
sonichi: But there are other newly added tests which are not skipped.

test/nlp/test_autohf_custom_metric.py
sonichi: I don't know why it causes key not found error. Optimizing eval_runtime doesn't look right either.

website/docs/Examples/AutoML-NLP.md
```diff
@@ -38,6 +38,13 @@ automl.fit(X_train=X_train, y_train=y_train, X_val=X_val, y_val=y_val, **automl_
 automl.predict(X_test)
 ```
 
+Notice that after you run `automl.fit`, the intermediate checkpoints are saved under `data/output`. You can use the following code to clean these outputs if they consume a large storage space:
```
sonichi, line 41:
```suggestion
Notice that after you run `automl.fit`, the intermediate checkpoints are saved under the specified output_dir `data/output`. You can use the following code to clean these outputs if they consume a large storage space:
```

website/docs/Examples/AutoML-NLP.md
```diff
@@ -38,6 +38,13 @@ automl.fit(X_train=X_train, y_train=y_train, X_val=X_val, y_val=y_val, **automl_
...
+
+```python
+if os.path.exists("test/data/output/"):
+    shutil.rmtree("test/data/output/")
```
sonichi, line 45:
```suggestion
if os.path.exists("data/output/"):
    shutil.rmtree("data/output/")
```

sonichi (approved):

qingyun-wu (approved):

PR #765, opened 2022-10-13
Docs

### LINKED ISSUE #759: Bug with implementation of SKLearnEstimator
I am currently using Flaml with the Sklearn regressor as described in the documentation. In particular I have 
``` python
class MyGBM(SKLearnEstimator):
    def __init__(self, task="regression", **config):
            super().__init__(task, **config)
            self.estimator_class = HistGradientBoostingRegressor

    @classmethod
    def search_space(cls, data_size, task):
        space = {
            "max_iter": {
                "domain": tune.randint(lower=100, upper=3000),
                "low_cost_init_value": 20,
            },
            "max_leaf_nodes": {
                "domain": tune.randint(lower=2, upper=2**8),
                "low_cost_init_value": 2,
            }
        }
        return space
```
However when I use this  I get
``` console
TypeError: HistGradientBoostingRegressor.__init__() got an unexpected keyword argument 'n_jobs'
```
which I think is a result of the signatures in the sklearn class not being able to handle this argument. 

I can solve this easily by creating my own class as follows

``` python
class MyHistGradientBoostingRegressor:
    def __init__(self, **kwargs):
        kwargs.pop("n_jobs", None)
        self.model = HistGradientBoostingRegressor(**kwargs)

    def fit(self, X, y, **kwargs):
        kwargs.pop("n_jobs", None)
        self.model.fit(X, y, **kwargs)
    
    def predict(self, X, **kwargs):
        return self.model.predict(X, **kwargs)
```
and then using this class instead. However it seems that either the documentation should be changed to reflect the need for the modlel to handle `kwargs` or the code should be altered. I can take a stab at fixing this if there is interest in solving the bug.

velezbeltran:

I added some changes that better explain what to do in the case where a Sklearn style estimator doesn't have an `n_jobs` argument. 

## Related issue number

Closes #759 

## Checks

- [x]I've made sure all auto checks have passed.

### DIFF
--- a/flaml/model.py
+++ b/flaml/model.py
@@ -924,8 +924,9 @@ class SKLearnEstimator(BaseEstimator):
     """
     The base class for tuning scikit-learn estimators.
 
-    Subclasses should accept a `n_jobs` parameter in their constructor
-    regardless of whether it is used by the estimator.
+    Subclasses should modify the function signature of ``__init__`` to
+    ignore the values in ``config`` that are not relevant to the constructor
+    of their estimator.
     """
 
     def __init__(self, task="binary", **config):
--- a/flaml/model.py
+++ b/flaml/model.py
@@ -921,7 +921,12 @@ class TransformersEstimatorModelSelection(TransformersEstimator):
 
 
 class SKLearnEstimator(BaseEstimator):
-    """The base class for tuning scikit-learn estimators."""
+    """
+    The base class for tuning scikit-learn estimators.
+
+    Subclasses should accept a `n_jobs` parameter in their constructor
+    regardless of whether it is used by the estimator.
+    """
 
     def __init__(self, task="binary", **config):
         super().__init__(task, **config)
--- a/website/docs/Use-Cases/Task-Oriented-AutoML.md
+++ b/website/docs/Use-Cases/Task-Oriented-AutoML.md
@@ -169,7 +169,7 @@ class MyRegularizedGreedyForest(SKLearnEstimator):
         return space
 ```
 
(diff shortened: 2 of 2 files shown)


sonichi:
@qingyun-wu relevant to tutorial.

flaml/model.py
```diff
@@ -921,7 +921,13 @@ def search_space(cls, data_size, task, **params):
 
 
 class SKLearnEstimator(BaseEstimator):
-    """The base class for tuning scikit-learn estimators."""
+    """
+    The base class for tuning scikit-learn estimators.
+
+    Subclasses should modify the function signature of ``__init__`` to
+    ignore the values in ``config`` that are not relevant to the constructor
+    of their estimator.
```
sonichi, line 929:
```suggestion
    Subclasses can modify the function signature of ``__init__`` to
    ignore the values in ``config`` that are not relevant to the constructor
    of their underlying estimator. For example, some regressors in ``scikit-learn``
    don't deal with the ``n_jobs`` parameter contained in ``config``. For those,
    one can add ``n_jobs=1,`` before ``**config`` to make sure ``config`` doesn't
    contain ``n_jobs``.
```
Is this more clear? I don't know the difference between ` and `` in docstr. Please make sure it looks good in the website.

sonichi (approved):

flaml/model.py
velezbeltran: Just update. I did check and it does look good 😊

ZviBaratz (commented):
Looks alright, I would just change the example to have `n_jobs=None` if possible (sorry if I'm missing something, haven't put it to the test). I'm assuming the estimator simply receives it as parts of `**kwargs` and it is not used. In that case, I find that declaring `n_jobs=1` is confusing, since it does assign it a particular value (rather than explicitly declaring there isn't one).

qingyun-wu (approved):

Issue #754, opened 2022-10-10
Time series gap detection for TFT tasks


gmdiana-hershey:
I've been getting a warning when I'm using TemporalFusionTransformer, indicating missing timestamps. It's coming from https://github.com/microsoft/FLAML/blob/f97c233dc55a60315f84fdb84fd49dd9ea98a5bd/flaml/automl.py#L1033. It Seems like perhaps it's checking the whole dataset as a single time series, rather than each time series individually. I think that adding logic to _validate_ts_data() to check if the task is 'ts_forecast_panel' and using the group_ids passed when fit() is called could be a solution.

int-chaos:
will fix!

PR #770, opened 2022-10-19
tft gap detection on each time series independently

int-chaos:
perform gap detection on each time series independently for panel time series datasets

#754

### DIFF
--- a/flaml/automl.py
+++ b/flaml/automl.py
@@ -1027,11 +1027,25 @@ class AutoML(BaseEstimator):
                 dataframe[[dataframe.columns[0]]].duplicated() is None
             ), "Duplicate timestamp values with different values for other columns."
         ts_series = pd.to_datetime(dataframe[dataframe.columns[0]])
-        inferred_freq = pd.infer_freq(ts_series)
-        if inferred_freq is None:
-            logger.warning(
-                "Missing timestamps detected. To avoid error with estimators, set estimator list to ['prophet']. "
-            )
+        if self._state.task == TS_FORECASTPANEL:
+            # check for each time series independently
+            group_ids = self._state.fit_kwargs.get("group_ids")
+            unique_ids = dataframe[group_ids].value_counts().reset_index()[group_ids]
+            for _, row in unique_ids:
+                df = dataframe.copy()
+                for id in group_ids:
+                    ts = df.loc[df[id] == row[id]]
+            inferred_freq = pd.infer_freq(ts)
+            if inferred_freq is None:
+                logger.warning(
+                    "Missing timestamps detected. To avoid error with estimators, set estimator list to ['prophet']. "
+                )
+        else:
+            inferred_freq = pd.infer_freq(ts_series)
+            if inferred_freq is None:
+                logger.warning(
+                    "Missing timestamps detected. To avoid error with estimators, set estimator list to ['prophet']. "
+                )
         if y_train_all is not None:
             return dataframe.iloc[:, :-1], dataframe.iloc[:, -1]
         return dataframe


flaml/automl.py
```diff
@@ -1026,12 +1026,27 @@ def _validate_ts_data(
...
+            for _, row in unique_ids.iterrows():
+                df = dataframe.copy()
+                for id in group_ids:
+                    ts = df.loc[df[id] == row[id]]
+            ts_series = pd.to_datetime(ts[ts.columns[0]])
+            inferred_freq = pd.infer_freq(ts_series)
+            if inferred_freq is None:
+                logger.warning(
+                    "Missing timestamps detected. To avoid error with estimators, set estimator list to ['prophet']. "
+                )
```
sonichi, line 1042: There are multiple problems:
1. Does prophet handle panel data? Does TFT handle missing timestamps? If TFT can handle missing timestamps, this check is not necessary. If prophet can't handle panel data, this message shouldn't suggest adding prophet.
2. For loop is not efficient. There should be a functional way using `groupby()`.
3. Do you intend to infer the frequency for each series? The current code only does it for the last one.

int-chaos: To clarify, TimeSeriesDataset only handle timestamps that are missing, but do not handle `NA` values.

int-chaos: @EgorKraevTransferwise and @markharley Our current solution to this issue will be to either require no missing timestamps or else user have to provide a `freq` argument.

EgorKraevTransferwise:
Here is how we handled it in our TFT wrapper, pivot by timestamp and dimension, add time index and fill any gaps, melt back into the original shape. Then it's enough that a timestamp is there for at least one of the series. We do use the frequency argument in there, but Pandas [can infer that](https://pandas.pydata.org/docs/reference/api/pandas.infer_freq.html).

```
def add_time_idx_new(data: BasicDataset, time_col: str) -> pd.DataFrame:

    pivoted = data.data.pivot(
        index=time_col,
        columns=data.metadata["dimensions"],
        values=data.metadata["metrics"],
    ).fillna(0.0)

    dt_index = pd.date_range(
        start=pivoted.index.min(),
        end=pivoted.index.max(),
        freq=data.metadata["frequency"],
    )
    indexes = pd.DataFrame(dt_index).reset_index()
    indexes.columns = ["time_idx", time_col]

    # this join is just to make sure that we get all the row timestamps in
    out = pd.merge(
        indexes, pivoted, left_on=time_col, right_index=True, how="left"
    ).fillna(0.0)

    # now flatten back

    melted = pd.melt(out, id_vars=[time_col, "time_idx"])

    for i, d in enumerate(["metric"] + data.metadata["dimensions"]):
        melted[d] = melted["variable"].apply(lambda x: x[i])

    # and finally, move metrics to separate columns
    re_pivoted = melted.pivot(
        index=["time_idx", time_col] + data.metadata["dimensions"],
        columns="metric",
        values="value",
    ).reset_index()

    return re_pivoted
```

int-chaos:
Understood. The current implementation uses `pd.infer_freq()` as well under the assumption that the user's data has no missing timestamps. However, the concern is that if there are missing timestamps, `pd.infer_freq` would return `None` and it won't be possible to use `pd.date_range`. We discussed this issue before but want to revisit it now since TFT handles missing timestamps and we want to leverage that ability.

So, in the case that there are missing timestamps, should we just tell the user to pass in a `freq` argument. This way we can still use `pd.date_range` to get the dates we need for the time series and create the `time_idx` column.

EgorKraevTransferwise:
I still think we should be guessing the frequency from the user input, even with missing data. This could work as follows:
1. get the ordered set of all distinct timestamps, so we'll only get a gap if there's a gap in *all* the time series
2. Calculate the diffs between adjacent timestamps
3. Get the median diff (most of them will be equal anyway so this should be quite robust)
4. Generate a new sequence of timestamps by starting at the first original timestamp and recursively adding the median diff to it, like 20 times
5. Apply `pd.infer_freq()` to that sequence

All in all, like 5 lines of code and should be quite robust

EgorKraevTransferwise:
"Imports are incorrectly sorted and/or formatted."  - why and how should I be sorting them? :)

int-chaos:
> I still think we should be guessing the frequency from the user input, even with missing data. This could work as follows:
> 
> 1. get the ordered set of all distinct timestamps, so we'll only get a gap if there's a gap in _all_ the time series
> 2. Calculate the diffs between adjacent timestamps
> 3. Get the median diff (most of them will be equal anyway so this should be quite robust)
> 4. Generate a new sequence of timestamps by starting at the first original timestamp and recursively adding the median diff to it, like 20 times
> 5. Apply `pd.infer_freq()` to that sequence
> 
> All in all, like 5 lines of code and should be quite robust
> 
> Or if you just want to use date_range, you can just take the sequence from 4. directly instead

```python
import pandas as pd
import numpy as np
import random
ts = list(pd.date_range("2017-01-01", periods=36, freq="M"))
ts.pop(random.randint(0, len(ts)-1))
diff = [ts[idx+1] - ts[idx] for idx in range(len(ts) - 1)]
med_diff = np.median(diff)
print(pd.date_range(start=ts[0], end=ts[-1]))
```

EgorKraevTransferwise:
Yes, it's a pretty small lookup table for the common ones, and those will cover 95% of all usecases; we should leave the option for the user to manually override the frequency instead, for the less common ones.

int-chaos:
> Yes, it's a pretty small lookup table for the common ones, and those will cover 95% of all usecases; we should leave the option for the user to manually override the frequency instead, for the less common ones.

got it!

Issue #730, opened 2022-09-13
Pickle Transformer


firmai:
It seems like I can't save the transformer object.
`
AttributeError: Can't pickle local object 'preprocess_and_suggest_hyperparams.<locals>.AutoMLTransformer'`

liususan091219:
Hi, can you share the code so we could reproduce the issue?

firmai:
It can be reproduced with any dataset, and then pickling the transformer object (feature_transformer)

```
        hyperparams, estimator_class, X_transformed, y_transformed, feature_transformer, label_transformer = preprocess_and_suggest_hyperparams(
            "classification", X, y, "lgbm"
        )
```

sonichi:
> It can be reproduced with any dataset, and then pickling the transformer object (feature_transformer)
> 
> ```
>         hyperparams, estimator_class, X_transformed, y_transformed, feature_transformer, label_transformer = preprocess_and_suggest_hyperparams(
>             "classification", X, y, "lgbm"
>         )
> ```

Thanks. There is an easy solution to fix this: move the definition of `AutoMLTransformer` out of the function and take `model` and `dt` as arguments in its constructor. Would you like to make a PR?

firmai:
Hi @sonichi thanks for the explanation, I am just getting acquainted with the library, so have decided to work around this issue for now.

PR #783, opened 2022-10-29
Fix issues related to zero-shot automl

### LINKED ISSUE #782: zero-shot config out of search space
For small datasets, the recommended zero-shot config may be of out of the search space because the search space depends on the dataset size. For example, `n_estimators` is bounded the # rows in the training data. The current behavior in `AutoML.fit()` is to remove the invalid configuration. If neither `time_budget` nor `max_iter` is given, we will end up training each learner with the low cost configuration. In this case, shall we ignore the search space constraint? Or shall we modify the configuration to fit into the search space?

sonichi:

- [x] For small datasets, the recommended zero-shot config may be of out of the search space because the search space depends on the dataset size. For example, n_estimators is bounded by the # rows in the training data but the zero-shot config could violate it. The current behavior in AutoML.fit() is to remove the invalid configuration. If neither time_budget nor max_iter is given, we will end up training each learner with the low cost configuration. This PR makes that validity check skipped when max_iter per learner is no larger than the number of init configs provided (hence no search is expected), by passing the right max_iter to each search state. Previously the total max_iter was passed. In this PR, we pass max_iter/#learners when roundrobin is used as the learner_selector. (#782)
- [x] When time_budget and max_iter are both unspecified in AutoML.fit(), only one trial per learner will be trained (using our [zero-shot](https://microsoft.github.io/FLAML/docs/Use-Cases/Zero-Shot-AutoML) technique). A zero-shot config can be an empty dict {} corresponding to an estimator's default. However, in AutoML.fit(), such a config will be then regarded as incomplete and modified. To prevent this, we need to update our approach of dealing with {} as the init config. We now set `init_config` as a list, potentially [] when no starting point is available. And we allow an empty dict to be used as is, by using an argument `allow_empty_config`. (#784)
- [x] Make AutoMLTransformer picklable (#730)

## Related issue number

Closes #782, #784, #730

## Checks


skzhang1 (approved):

flaml/automl.py
```diff
@@ -183,20 +183,24 @@ def __init__(
                 isinstance(starting_point, dict)
                 and starting_point.get(name) is not None
             ):
+                if self.init_config is None:
+                    self.init_config = {}
```
ZviBaratz, line 187: Would it be cleaner to move the evaluation of `self.init_config` to before the for-loop? Looks like doing it once before starting the iteration would have the same result.

ZviBaratz (approved):

flaml/automl.py
sonichi: There is a chance that the evaluation of `self.init_config` would not be evaluated at all in the for-loop. If we move it out, we would be setting self.init_config to {} no matter what. That's what I wanted to avoid.

ZviBaratz: I see. Feels like the logic could benefit from some refactoring. However, that may very well be best kept to a separate PR.

qingyun-wu (approved):

PR #833, opened 2022-11-30
catch TFT logger bugs

int-chaos:
catch errors with pl.Trainer loggers for TemporalFusionTransformer

### DIFF
--- a/flaml/model.py
+++ b/flaml/model.py
@@ -2325,7 +2325,7 @@ class TemporalFusionTransformerEstimator(SKLearnEstimator):
                 kwargs.get("log_dir", "lightning_logs")
             )  # logging results to a tensorboard
             trainer = _fit(log=logger)
-        except ValueError:
+        except (ValueError, AttributeError):
             # issue with pytorch forecasting model log_prediction() function
             # pytorch-forecasting issue #1145
             trainer = _fit(log=False)
--- a/flaml/model.py
+++ b/flaml/model.py
@@ -2325,7 +2325,7 @@ class TemporalFusionTransformerEstimator(SKLearnEstimator):
                 kwargs.get("log_dir", "lightning_logs")
             )  # logging results to a tensorboard
             trainer = _fit(log=logger)
-        except:
+        except ValueError:
             trainer = _fit(log=False)
         best_model_path = trainer.checkpoint_callback.best_model_path
         best_tft = TemporalFusionTransformer.load_from_checkpoint(best_model_path)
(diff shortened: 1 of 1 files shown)


flaml/model.py
```diff
@@ -2287,36 +2288,45 @@ def fit(self, X_train, y_train, budget=None, **kwargs):
...
-            reduce_on_plateau_patience=4,
-        )
-        # fit network
-        trainer.fit(
-            tft,
-            train_dataloaders=train_dataloader,
-            val_dataloaders=val_dataloader,
-        )
+
+        def _fit(log):
```
sonichi, line 2292: Can you move the definition out of this method? It will make this class unpicklable. That will result in an error when `use_ray=True` or when the AutoML object is pickled. Please also add a pickle statement in your test, similar to some other tests under automl/.

flaml/model.py
```diff
@@ -2287,36 +2288,45 @@ def fit(self, X_train, y_train, budget=None, **kwargs):
...
+                val_dataloaders=val_dataloader,
+            )
+            return trainer
+
+        try:
+            logger = TensorBoardLogger(
+                kwargs.get("log_dir", "lightning_logs")
+            )  # logging results to a tensorboard
+            trainer = _fit(log=logger)
+        except:
```
sonichi, line 2328: Need to add a specific Exception class after `except`. Also add comment why to do this and the issue number to track in pytorch-forecast. In future we may remove it when the issue is fixed in pytorch-forecast.

flaml/model.py
sonichi: You already have pickle in your test. So please ignore this comment.

sonichi (approved):

flaml/model.py
```diff
@@ -2287,36 +2288,47 @@ def fit(self, X_train, y_train, budget=None, **kwargs):
...
+
+        try:
+            logger = TensorBoardLogger(
+                kwargs.get("log_dir", "lightning_logs")
+            )  # logging results to a tensorboard
+            trainer = _fit(log=logger)
+        except (ValueError, AttributeError):
+            # issue with pytorch forecasting model log_prediction() function
+            # pytorch-forecasting issue #1145
+            trainer = _fit(log=False)
```
sonichi, line 2331: AttributeError was raised in this line instead of line 2327. I guess that means `log=False` doesn't work.
If you confirm that, and can't find a way to make it work, then this line should be removed. Instead, issue a warning that an error occurred in pytorch-forecasting. In `.predict()`, make a dumb prediction for such cases.

int-chaos: issue was had to set `callbacks=False` as well

qingyun-wu (approved):

PR #836, opened 2022-12-03
disable tft logger

int-chaos:
#833 did not fix issue. new attempt to disable logger by adding set `log_interval=0` when `log=False`


sonichi (approved):

sonichi (commented):
Is early_stop_callback invalid when log=False?

sonichi:
Why don't we disable logging until the issue in pytorch-forecasting is fixed?

sonichi:
> Why don't we disable logging until the issue in pytorch-forecasting is fixed?

I did that in #837

int-chaos:
> > Why don't we disable logging until the issue in pytorch-forecasting is fixed?
> 
> I did that in #837

sounds good, will close PR

PR #846, opened 2022-12-08
Add supporting using Spark as the backend of parallel training

thinkall:

FLAML doesn't support using spark as the backend for parallel training, but many data scientists are processing big data and building machine learning models with Spark cluster. As an AutoML library, FLAML would benefit broader users by supporting Spark.

## Checks


flaml/automl/automl.py
```diff
@@ -2584,6 +2668,10 @@ def cv_score_agg_func(val_loss_folds, log_metrics_folds):
                 X_train = ray.get(X_train)
             elif isinstance(dataframe, ray.ObjectRef):
                 dataframe = ray.get(dataframe)
+        else:
+            self._state.resources_per_trial = (
+                {"cpu": n_jobs} if n_jobs > 0 else {"cpu": 1}
+            )
```
sonichi, line 2674: Is this correct? Add "TODO" if you envision this to be changed in future.

flaml/automl/automl.py
```diff
@@ -2358,6 +2401,10 @@ def custom_metric(
                 datasets, but will incur more overhead in time.
                 If dict: the dict contains the keywords arguments to be passed to
                 [ray.tune.run](https://docs.ray.io/en/latest/tune/api_docs/execution.html).
+            use_spark: boolean, default=False | Whether to use spark to run the training
+                in parallel spark jobs. This can be used to accelerate training on large models
+                and large datasets, but will incur more overhead in time and thus slow down
+                training in some cases.
```
sonichi, line 2407: Shall we mention that gpu is not supported when use_spark=True?

flaml/automl/automl.py
```diff
@@ -3550,6 +3659,10 @@ def _search(self):
                         and ray.available_resources()["CPU"]
                         or os.cpu_count()
                     )
+                elif self._use_spark is not False:
+                    from flaml.utils import get_n_cpus
+
+                    n_cpus = get_n_cpus()
```
sonichi, line 3665: Do you have a test that covers this line? That is, both use_spark and ensemble=True.
Also, here the ensemble is run in the driver node. But `get_n_cpus()` returns the number of cpus in worker nodes.
Consider doing the ensemble in worker nodes in future?

sonichi (commented):
Other docs to update:
https://microsoft.github.io/FLAML/docs/Use-Cases/Task-Oriented-AutoML#parallel-tuning
https://microsoft.github.io/FLAML/docs/Use-Cases/Tune-User-Defined-Function#parallel-tuning
In future, consider adding an example similar to
https://microsoft.github.io/FLAML/docs/Examples/Integrate%20-%20AzureML

Dockerfile
```diff
@@ -3,6 +3,16 @@ FROM python:3.7
 RUN apt-get update && apt-get -y update
```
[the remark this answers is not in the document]
thinkall: Yes, I've built a docker image with the updated Dockerfile, and then started a container with the image.

In the container, I downloaded my code and installed FLAML[test] and ran `pytest test/spark`, all tests passed.

flaml/tune/trial_runner.py
```diff
@@ -135,3 +135,44 @@ def step(self) -> Trial:
...
+        else:
+            trial = None
+        return trial
+
+    def stop_trial(self, trial):
+        super().stop_trial(trial)
+        try:
+            self.running_trials.remove(trial)
+        except ValueError:
+            logger.warning("Trial {} is not running.".format(trial.trial_id))
```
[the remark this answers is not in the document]
thinkall: It happens when a user calls runner.stop(trial) manually, but the trial is not in the runner.running_trials. Then remove an item not in a list will raise an ValueError. It should never happen if user's code is valid, I added this just for robust.

flaml/tune/tune.py
```diff
@@ -359,7 +362,7 @@ def easy_objective(config):
         print(analysis.trials[-1].last_result)
     ```
 
-        verbose: 0, 1, 2, or 3. Verbosity mode for ray if ray backend is used.
+        verbose: 0, 1, 2, or 3. Verbosity mode for ray and spark if backend is used.
```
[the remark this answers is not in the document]
thinkall: Ah, not exactly the same, Spark has different number of levels for logs. I'll recalibrate it.

setup.py
```diff
@@ -44,6 +44,11 @@
             "matplotlib",
             "openml==0.10.2",
         ],
+        "spark": [
+            "pyspark>=3.0.0",
+            "optuna==2.8.0",
```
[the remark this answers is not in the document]
thinkall: But the default case I tried failed without optuna. This is why I added it.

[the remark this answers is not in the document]
sonichi: That is unexpected. It shouldn't fail unless the hpo method is "bs".

Dockerfile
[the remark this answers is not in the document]
sonichi: Great. Thanks.

flaml/automl/automl.py
thinkall: It's supported I think. Need more tests though.

setup.py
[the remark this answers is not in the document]
thinkall: Ah, yes. I think I mixed spark and our platform requirements a little bit.

flaml/automl/automl.py
sonichi: There is an argument "gpu_per_trial" which is related to this. If gpu is supported, lines involving it would need to be checked and adapted.

test/spark/test_performance.py
```diff
@@ -0,0 +1,117 @@
...
+    if budget is None:
+        assert accuracy >= 0.669, "the accuracy of flaml should be larger than 0.67"
+
+
+@pytest.mark.skipif(skip_spark, reason="Spark is not installed. Skip all spark tests.")
+def test_automl_array():
+    run_automl(5, "array", "bs")
+
+
+@pytest.mark.skipif(skip_spark, reason="Spark is not installed. Skip all spark tests.")
```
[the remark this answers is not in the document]
thinkall: Test only with linux+python310 now, and set time_budget to 3600, otherwise, it can take more than 2 hours as I tried  today.

In test/automl/test_notebook_example.py, no performance test will be triggered.

flaml/tune/trial_runner.py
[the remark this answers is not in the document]
sonichi: This is not a user-facing class. It's developer-facing. If this happens then it means there is a bug in developer's code. An assertion is a better fit than a warning message.

flaml/tune/tune.py
```diff
@@ -605,6 +612,136 @@ def easy_objective(config):
...
+                                            # When the result returned is an empty dict, set the trial status to error
+                                            trial_to_run.set_status(Trial.ERROR)
+                                    else:
+                                        report(_metric=result)
+                                _runner.stop_trial(trial_to_run)
+                            fail = 0
+                    if fail >= ub:
+                        logger.warning(
+                            f"fail to sample a trial for {max_failure} times in a row, stopping."
+                        )
```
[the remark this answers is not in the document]
sonichi: That's why I suggest changing the condition to be `not trials_to_run`. If `trials_to_run` is empty we should break the loop after line 694. And we don't need to check `fail>=ub` here.

qingyun-wu:
There are several places with expressions like `if self._use_spark is not False`. Any specific reason why it cannot be simplified to `if self._use_spark` ?

flaml/tune/tune.py
```diff
@@ -605,6 +612,136 @@ def easy_objective(config):
...
+
+        register_spark()
+        spark = SparkSession.builder.getOrCreate()
+        sc = spark._jsc.sc()
+        num_executors = (
+            len([executor.host() for executor in sc.statusTracker().getExecutorInfos()])
+            - 1
+        )
+        """
+        By default, the number of executors is the number of VMs in the cluster. And we can
```
qingyun-wu, line 640: This chunk of docstr and the setting in line 662 of this file are quite informative and important. Should we make it more visible? For example, adding instructions or a reference in the automl.py file when [Spark is enabled](https://github.com/microsoft/FLAML/blob/62cd35322ecd90eef9d81f6318d0aa53512e652d/flaml/automl/automl.py#L688)

flaml/tune/tune.py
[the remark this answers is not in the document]
thinkall: Get it, thanks.

flaml/tune/trial_runner.py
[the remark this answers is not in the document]
thinkall: Ok. Maybe I can do the remove and let the original exception raise.

flaml/utils.py
```diff
@@ -0,0 +1,170 @@
...
+            SparkSession.builder.getOrCreate()
+            .sparkContext.getConf()
+            .get("spark.executor.cores")
+        )
+    except Exception:
+        n_cpus = os.cpu_count()
+    return n_cpus
+
+
+def with_parameters(trainable, **kwargs):
```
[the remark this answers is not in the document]
thinkall: I moved utils.py to folder flaml/spark

flaml/automl/automl.py
thinkall: It's not used in spark for now. Added TODO.

flaml/automl/automl.py
thinkall: resources_per_trial is not used in Sequential and Spark for now, but Spark cluster supports GPU, so I believe it can be supported in flaml[spark].

I added gpu is not supported when use_spark=True since we're now focusing on CPU clusters.

flaml/automl/automl.py
thinkall: Test added, get_n_cpus() updated.

thinkall:
> There are several places with expressions like `if self._use_spark is not False`. Any specific reason why it cannot be simplified to `if self._use_spark` ?

For consistent with use_ray  at the begining, but  it's actually unnecessary. All simplified now. Thanks.

flaml/tune/tune.py
thinkall: Thanks for the suggestion. Docs updated.

notebook/integrate_spark.ipynb
```diff
@@ -0,0 +1 @@
...ore\n","print('r2', '=', 1 - sklearn_metric_loss_score('r2', y_pred, y_test))\n","print('mse', '=', sklearn_metric_loss_score('mse', y_pred, y_test))\n","print('mae', '=', sklearn_metric_loss_score('mae', y_pred, y_test))"]},{"cell_type":"code","execution_count":null,"metadata":{"jupyter":{"outputs_hidden":false,"source_hidden":false},"nteract":{"transient":{"deleting":false}}},"outputs":[],"source":[]}],"metadata":{"kernel_info":{"name":"synapse_pyspark"},"kernelspec":{"display_name":"Python 3.8.13 ('syml-py38')","language":"python","name":"python3"},"language_info":{"codemirror_mode":{"name":"ipython","version":3},"file_extension":".py","mimetype":"text/x-python","name":"python","nbconvert_exporter":"python","pygments_lexer":"ipython3","version":"3.8.13"},"notebook_environment":{},"save_output":true,"spark_compute":{"compute_id":"/trident/default","session_options":{"conf":{"spark.livy.synapse.ipythonInterpreter.enabled":"true"},"enableDebugMode":false,"keepAliveTimeout":30}},"synapse_widget":{"state":{},"version":"0.1"},"trident":{"lakehouse":{}},"vscode":{"interpreter":{"hash":"e3d9487e2ef008ade0db1bc293d3206d35cb2b6081faff9f66b40b257b7398f7"}}},"nbformat":4,"nbformat_minor":0}
```
sonichi, line 1: To run this notebook example, please install flaml with the `notebook` option:\n","```bash\n","pip install flaml[spark,notebook]\n
->
To run this notebook example, please install flaml with the following options:\n"

sonichi (commented):
The beginning part of the spark notebook can be double checked.

website/docs/Use-Cases/Task-Oriented-AutoML.md
```diff
@@ -382,7 +382,11 @@ and have ``split`` and ``get_n_splits`` methods with the same signatures.  To di
 
 When you have parallel resources, you can either spend them in training and keep the model search sequential, or perform parallel search. Following scikit-learn, the parameter `n_jobs` specifies how many CPU cores to use for each training job. The number of parallel trials is specified via the parameter `n_concurrent_trials`. By default, `n_jobs=-1, n_concurrent_trials=1`. That is, all the CPU cores (in a single compute node) are used for training a single model and the search is sequential. When you have more resources than what each single training job needs, you can consider increasing `n_concurrent_trials`.
 
-To do parallel tuning, install the `ray` and `blendsearch` options:
+FLAML now support two backends for parallel tuning, i.e., `Ray` and `Spark`. You can use either of them, but not both for one tuning.
```
sonichi, line 385:
```suggestion
FLAML now support two backends for parallel tuning, i.e., `Ray` and `Spark`. You can use either of them, but not both for one tuning job.
```

website/docs/Installation.md
```diff
@@ -50,6 +50,28 @@ pip install flaml[nlp]
 ```bash
 pip install flaml[ray]
 ```
+* spark
+> *Spark supporting is added in v1.1.0*
```
sonichi, line 54:
```suggestion
> *Spark support is added in v1.1.0*
```

website/docs/Use-Cases/Task-Oriented-AutoML.md
```diff
@@ -397,6 +401,23 @@ automl.fit(X_train, y_train, n_jobs=4, n_concurrent_trials=4)
 ```
 flaml will perform 4 trials in parallel, each consuming 4 CPU cores. The parallel tuning uses the [BlendSearch](Tune-User-Defined-Function##blendsearch-economical-hyperparameter-optimization-with-blended-search-strategy) algorithm.
 
+#### Parallel tuning with Spark
+
+To do parallel tuning with Spark, install the `spark` and `blendsearch` options:
+
+> *Spark supporting is added in v1.1.0*
```
sonichi, line 408:
```suggestion
> *Spark support is added in v1.1.0*
```

notebook/integrate_spark.ipynb
```diff
@@ -0,0 +1 @@
...
```
sonichi, line 1: spark supporting -> spark support

sonichi (commented):
There are a number of lint issues you can see from the lint step in build. 
[pre-commit](https://microsoft.github.io/FLAML/docs/Contribute#pre-commit) can help it.

flaml/tune/tune.py
```diff
@@ -605,6 +612,141 @@ def easy_objective(config):
...
+                                    else:
+                                        # When the result returned is an empty dict, set the trial status to error
+                                        trial_to_run.set_status(Trial.ERROR)
+                                else:
+                                    logger.info(
+                                        "Brief result: {}".format({metric: result})
+                                    )
+                                    report(_metric=result)
+                            _runner.stop_trial(trial_to_run)
+                        fail = 0
```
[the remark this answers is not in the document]
thinkall: You mean `fail = 0`?

I think it's needed. By set fail to 0 here, we count `consecutive failures`; otherwise, we count `accumulative faiures`.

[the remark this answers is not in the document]
sonichi: It's redundant with line 677.

test/spark/test_ensemble.py
```diff
@@ -0,0 +1,58 @@
...
+        from flaml.spark.mylearner import MyRegularizedGreedyForest
+
+        skip_my_learner = False
+    except ImportError:
+        skip_my_learner = True
+else:
+    skip_my_learner = True
+
+
+class TestMultiClass(unittest.TestCase):
```
sonichi, line 29: TestEnsemble?

sonichi (commented):
lint issue: https://github.com/microsoft/FLAML/actions/runs/3702763945/jobs/6273408939#step:11:15

sonichi (commented):
lint issue: https://github.com/microsoft/FLAML/actions/runs/3702763945/jobs/6273408939#step:11:15

flaml/tune/tune.py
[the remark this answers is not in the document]
thinkall: Line 677 is to init `fail`, but line 731 is to reset `fail` for the while loop.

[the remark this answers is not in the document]
sonichi: My bad. You are right.

sonichi (commented):
I have no further comments now. Waiting for others' input on the top-level package question.

flaml/spark/__init__.py
```diff
@@ -0,0 +1,8 @@
+from flaml.spark.utils import (
```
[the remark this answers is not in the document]
qingyun-wu: Adding this top-level package seems not necessary, at least for now, to me. I am OK with this if there are reasonable justifications or if anyone has a strong opinion about this. What do you think @markharley?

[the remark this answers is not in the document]
qingyun-wu: A more concrete suggestion for the top-level spark module: it is perhaps better to move it to the `flaml.tune` folder （in this case, the functions in the [utils.py](https://github.com/microsoft/FLAML/pull/846/files/138447b3eb02eed22e65b7561ed18478a60a19c5#diff-9a80e3fdd93858985f28b172646435385c50037c928a7c84badefe3f3828aebb), e.g., `customize_learner` shall be revised into a more general one. ) 
But this potential change could be considered in a later PR.

qingyun-wu:
> I have no further comments now. Waiting for others' input on the top-level package question.

I added my comments about the top-level package question and created an issue for further follow-up discussion.

qingyun-wu (approved):

sonichi (approved):
I agree with Qingyun's comments. @thinkall If you agree too, I can merge the PR.

flaml/automl/automl.py
```diff
@@ -54,17 +55,36 @@
...
 try:
     import mlflow
 except ImportError:
     mlflow = None
 
+try:
+    import pyspark
+    from pyspark.sql import SparkSession
+
+    spark_import = True
```
markharley, line 75: A better name for these might be `spark_available` and `ray_available`

flaml/automl/automl.py
```diff
@@ -331,7 +351,7 @@ def _prepare_sample_train_data(self, sample_size):
         return sampled_X_train, sampled_y_train, sampled_weight, groups
 
     @staticmethod
-    def _compute_with_config_base(config_w_resource, state, estimator):
+    def _compute_with_config_base(config_w_resource, state, estimator, is_report=True):
```
markharley, line 354: I don't think `is_report` makes sense here. It's not that something is or isn't a report, but rather that if we're using ray then we need to report the score back to tune. Might be better just to pass `use_ray`

flaml/automl/automl.py
```diff
@@ -2081,8 +2117,13 @@ def trainable(self) -> Callable[[dict], Optional[float]]:
...
         mem_res = self._mem_thres
 
-        def train(config: dict, state):
-
+        def train(config: dict, state, is_report=True):
+            # handle spark broadcast variables
+            if spark_import:
+                if isinstance(state, pyspark.broadcast.Broadcast):
+                    state = state.value
+                if isinstance(is_report, pyspark.broadcast.Broadcast):
```
markharley, line 2125: Why would `is_report` ever by this type? Seems it should always be boolean. (Also see comment on the naming of this above)

flaml/automl/automl.py
```diff
@@ -2081,8 +2117,13 @@ def trainable(self) -> Callable[[dict], Optional[float]]:
...
+                    is_report = is_report.value
```
markharley, line 2126: Might be nice to put this logic in a function in the new spark utils module

flaml/automl/automl.py
```diff
@@ -2560,12 +2614,49 @@ def cv_score_agg_func(val_loss_folds, log_metrics_folds):
...
+                    "Spark installed, setting use_spark to True. If you want to use Ray, set use_ray to True."
+                )
+                use_spark = True
+                check_spark()
+            else:
+                logger.warning(
+                    "n_concurrent_trials > 1 is only supported when using Ray or Spark. "
+                    "Neither Ray nor Spark installed, setting n_concurrent_trials to 1."
+                )
+                n_concurrent_trials = 1
```
markharley, line 2651: Do we just want to throw here? Feels like misconfiguration to ask for concurrent trials with no parallel executor

flaml/spark/__init__.py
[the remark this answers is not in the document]
markharley: I agree @qingyun-wu, it seems best to move this to the `flaml.tune` subpackage

flaml/spark/utils.py
```diff
@@ -0,0 +1,173 @@
...
+    flaml_path = os.path.dirname(os.path.abspath(__file__))
+    learner_code = textwrap.dedent(learner_code)
+    learner_path = os.path.join(flaml_path, file_name + ".py")
+
+    def writefile(_):
+        with open(learner_path, "w") as f:
+            f.write(learner_code)
+
+    # write learner into a file
+    writefile(None)
```
markharley, line 172: Would be neater to inline this. Don't feel the nested function is adding anything and odd to have an unused argument

flaml/tune/tune.py
```diff
@@ -605,6 +613,141 @@ def easy_objective(config):
...
+                        search_alg=search_alg,
+                        scheduler=scheduler,
+                        metric=metric,
+                        mode=mode,
+                    )
+                    num_trials = 0
+                    if time_budget_s is None:
+                        time_budget_s = np.inf
+                    fail = 0
+                    ub = (
```
markharley, line 678: What's `ub`?

flaml/tune/tune.py
```diff
@@ -605,6 +613,141 @@ def easy_objective(config):
...
+            max_concurrent = max(1, search_alg.max_concurrent)
+        else:
+            max_concurrent = max(1, int(os.getenv("FLAML_MAX_CONCURRENT", 1)))
+
+        n_concurrent_trials = min(num_executors, max_concurrent)
+        with parallel_backend("spark"):
+            with Parallel(
+                n_jobs=n_concurrent_trials, verbose=max(0, (verbose - 1) * 50)
+            ) as parallel:
+                try:
```
markharley, line 667: Would be nice to extract the internal of this try-except block to a function so that we're not so nested in context

flaml/tune/tune.py
```diff
@@ -605,6 +613,141 @@ def easy_objective(config):
...
+                    _runner = SparkTrialRunner(
+                        search_alg=search_alg,
+                        scheduler=scheduler,
+                        metric=metric,
+                        mode=mode,
+                    )
+                    num_trials = 0
+                    if time_budget_s is None:
+                        time_budget_s = np.inf
+                    fail = 0
```
markharley, line 677: Is this the count `num_failures`?

setup.py
```diff
@@ -44,6 +44,10 @@
             "matplotlib",
             "openml==0.10.2",
         ],
+        "spark": [
+            "pyspark>=3.0.0",
+            "joblibspark",
```
markharley, line 49: We should add some version constraints here so we don't get caught out on major upgrades

flaml/spark/utils.py
```diff
@@ -0,0 +1,173 @@
...
+    for k, v in kwargs.items():
+        bc_kwargs[k] = spark.sparkContext.broadcast(v)
+
+    return partial(trainable, **bc_kwargs)
+
+
+def customize_learner(learner_code="", file_name="mylearner"):
+    """Write customized learner/metric code contents to a file for importing.
+    It is necessary for using the customized learner in spark backend.
+    The path of the learner file will be returned.
```
markharley, line 124: Don't really get the purpose of this. Why would we write code to a string and write it back out again?

flaml/tune/tune.py
```diff
@@ -605,6 +613,141 @@ def easy_objective(config):
...
+                        logger.debug(
+                            f"Configs of Trials to run: {[trial_to_run.config for trial_to_run in trials_to_run]}"
+                        )
+                        results = parallel(
+                            delayed(evaluation_function)(trial_to_run.config)
+                            for trial_to_run in trials_to_run
+                        )
+                        # results = [evaluation_function(trial_to_run.config) for trial_to_run in trials_to_run]
+                        while results:
+                            result = results.pop(0)
```
markharley, line 715: Why not just iterate over them?

flaml/tune/tune.py
```diff
@@ -605,6 +613,141 @@ def easy_objective(config):
...
+                            trial_to_run = trials_to_run[0]
```
markharley, line 716: Do we want this to be always the first entry?

test/spark/test_performance.py
```diff
@@ -0,0 +1,112 @@
...
+    print("accuracy", "=", accuracy)
+    print(
+        "roc_auc", "=", 1 - sklearn_metric_loss_score("roc_auc", y_pred_proba, y_test)
+    )
+    print("log_loss", "=", sklearn_metric_loss_score("log_loss", y_pred_proba, y_test))
+    if performance_check_budget is None:
+        assert accuracy >= 0.669, "the accuracy of flaml should be larger than 0.67"
+
+
+@pytest.mark.skipif(skip_spark, reason="Spark is not installed. Skip all spark tests.")
```
markharley, line 100: To skip all the tests of a module you can use
```
pytestmark = pytest.mark.skipif(...)
```
at the top level of the module. Saves duplicating the decorator

markharley: https://docs.pytest.org/en/7.1.x/how-to/skipping.html#skip-all-test-functions-of-a-class-or-module

flaml/tune/tune.py
thinkall: We wanted to make the compute of results to be async, in that case, results can be changed during the processing of result. The async mode may be added in a future PR.

flaml/tune/tune.py
thinkall: result and trial_to_run are one-to-one mappings, here we always get the first items of both results and trials_to_run to keep the right mapping.

test/spark/test_performance.py
thinkall: Thank you @markharley , this is great!

flaml/spark/utils.py
thinkall: If a class is defined inline, it can't be passed to spark executors correctly, we need to import it from a file/module. This is a limitation of pyspark.

flaml/tune/tune.py
thinkall: Yes, the number of consecutive failures. The naming here is consistent with that in sequential mode.

flaml/tune/tune.py
thinkall: @sonichi could you help with this one? Thanks.

flaml/automl/automl.py
thinkall: The reason I just log a warning here is to make flaml more friendly to users. The same config can run in different environments.

flaml/automl/automl.py
thinkall: I think `spark_import` and `ray_import` is more accurate. As this import code will pass if pyspark or ray is installed. However, a real spark cluster or ray cluster is not yet set up. That's why we have `check_spark` in spark.utils.

flaml/automl/automl.py
thinkall: Maybe we will implement async report for Spark too. In that case, would `is_report` make more sense? Plus, `use_ray` kinds of limiting things to `ray` or `not ray`, but is the fact here more about `report` or `not report`?

flaml/tune/tune.py
thinkall: I hope to keep it this way as including a new function will change the scope of variables which might cause issues.

flaml/tune/tune.py
sonichi: upper bound. Sorry for the naming and not having a comment in the code.

flaml/tune/spark/__init__.py
```diff
@@ -0,0 +1,8 @@
+from flaml.tune.spark.utils import (
+    check_spark,
+    get_n_cpus,
+    with_parameters,
+    customize_learner,
```
sonichi, line 5: `customize_learner` is a concept in `automl`. There are two ways to deal with it.
1. Rename `customize_learner` into a more general concept, such as `broadcast_code` which is suitable for the `tune` layer.
2. Move `customize_learner` into `flaml.automl`.

I prefer 1 because even without doing automl, there can be cases where the broadcasting is needed when using `flaml.tune`.

sonichi (commented):
I have one comment. I'm OK with addressing it in the next PR. If you plan to do it in the next PR, could you please create an issue to track it?

flaml/tune/spark/__init__.py
thinkall: Refactored to broadcast_code. Thanks.

flaml/tune/tune.py
sonichi: #858 created for renaming it in future PRs.

flaml/tune/tune.py
sonichi: #859 created for addressing it in future PR.

flaml/automl/automl.py
sonichi: #860 created in case we want to change the behavior in future.

flaml/tune/tune.py
sonichi: #861 created to keep track of this suggestion.

PR #848, opened 2022-12-10
fix bug related to _choice_

sonichi:

1. Before a trial is executed in AutoML, `_choice_` needs to be popped out of the config. Otherwise the constructor of an estimator could fail due to unrecognized parameter. In several other places, the config should be sanitized too.
2. In the search algorithm, the key `_choice_` may not exist in the dict when it's accessed. Using "get" to avoid KeyError.
3. Removed py3.6 from the build as the build stops working for py3.6.

## Checks

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-2019]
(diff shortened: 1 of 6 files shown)


thinkall (approved):
Verified with spark.

skzhang1 (approved):

qingyun-wu (approved):

PR #857, opened 2022-12-20
Extract task class from automl

markharley:
This is the next PR in the series of refactors. In this we introduce the Task class to begin abstracting the task types. At present, we only introduce the abstract base Task class and a GenericTask which contains the task specific logic that was contained in AutoML.

## Related issue number
None

## Checks

- [X] I've used [pre-commit](https://microsoft.github.io/FLAML/docs/Contribute#pre-commit) to lint the changes in this PR, or I've made sure [lint with flake8](https://github.com/microsoft/FLAML/blob/816a82a1155b4de4705b21a615ccdff67c6da379/.github/workflows/python-package.yml#L54-L59) output is two 0s.


flaml/automl/task/task.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def __init__(
+        self,
+        task_name: str,
+        X_train: Optional[Union[np.ndarray, pd.DataFrame]],
+        y_train: Optional[Union[np.ndarray, pd.DataFrame, pd.Series]],
+    ):
+        self.name = task_name
+        if X_train is not None:
+            self.train_data_size = X_train.shape[0]
```
sonichi, line 67: Where is this field used?

markharley:
Hey! This should be ready to review now. Looks like we got the last tests fixed up

flaml/automl/ml.py
```diff
@@ -697,6 +593,7 @@ def train_estimator(
     return estimator, train_time
 
 
+# TODO(Mark) Can remove?
```
sonichi, line 596: It seems to be moved to task.py.

test/automl/test_multiclass.py
```diff
@@ -3,8 +3,9 @@
 import scipy.sparse
 from sklearn.datasets import load_iris, load_wine
 from flaml import AutoML
-from flaml.automl.data import CLASSIFICATION, get_output_from_log
+from flaml.automl.data import get_output_from_log
 from flaml.automl.model import LGBMEstimator, XGBoostSklearnEstimator, SKLearnEstimator
+from flaml.automl.task.task import CLASSIFICATION
```
sonichi, line 8: Is line 18 still correct when task is a str?

test/spark/test_multiclass.py
```diff
@@ -3,7 +3,8 @@
 import scipy.sparse
 from sklearn.datasets import load_iris, load_wine
 from flaml import AutoML
-from flaml.automl.data import CLASSIFICATION, get_output_from_log
+from flaml.automl.data import get_output_from_log
+from flaml.automl.task.task import CLASSIFICATION
```
sonichi, line 7: It seems not used.

markharley: Thanks! Removed

test/automl/test_multiclass.py
markharley: I've fixed it now, thanks

notebook/automl_classification.ipynb
```diff
@@ -858,7 +858,7 @@
     "''' SKLearnEstimator is the super class for a sklearn learner '''\n",
     "from flaml.model import SKLearnEstimator\n",
     "from flaml import tune\n",
-    "from flaml.data import CLASSIFICATION\n",
+    "from flaml.automl.task.task import CLASSIFICATION\n",
```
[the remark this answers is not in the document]
markharley: Fixed, thanks

flaml/automl/ml.py
markharley: Good spot, removed this one now

flaml/automl/model.py
```diff
@@ -2031,7 +2040,7 @@ def search_space(cls, data_size, pred_horizon, **params):
         space.update(
             {
                 "optimize_for_horizon": {
-                    "domain": tune.choice([True, False]),
+                    "domain": tune.choice([False]), #tune.choice([True, False]),
```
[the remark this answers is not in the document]
markharley: Not sure where this came from... possibly a bad merge. Changed back

flaml/automl/task/task.py
```diff
@@ -0,0 +1,199 @@
...
+
+    def __init__(
+        self,
+        task_name: str,
+        X_train: Optional[Union[np.ndarray, pd.DataFrame]],
+        y_train: Optional[Union[np.ndarray, pd.DataFrame, pd.Series]],
+    ):
+        self.name = task_name
+        if ray is not None and isinstance(X_train, ray.ObjectRef):
+            X_train = ray.get(X_train)
```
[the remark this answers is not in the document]
markharley: I think this was a bad merge again. Removed

flaml/automl/automl.py
```diff
@@ -1757,15 +776,19 @@ def retrain_from_log(
...
+            task = task_factory(task)
+
         eval_method = eval_method or self._settings.get("eval_method")
         split_ratio = split_ratio or self._settings.get("split_ratio")
         n_splits = n_splits or self._settings.get("n_splits")
         split_type = split_type or self._settings.get("split_type")
         auto_augment = (
             self._settings.get("auto_augment") if auto_augment is None else auto_augment
         )
+        self.task = task
```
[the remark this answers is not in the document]
markharley: I've just replaced it with `self._state.task` for now. Don't think we need the property necessarily since it's only internal

markharley:
Fixed the lint issues too, thanks

notebook/automl_classification.ipynb
```diff
@@ -875,7 +875,10 @@
     "        super().__init__(task, **config)\n",
     "\n",
     "        '''task=binary or multi for classification task'''\n",
-    "        if task in CLASSIFICATION:\n",
+    "        if isinstance(task, str):\n",
+    "            from flaml.automl.task.factory import task_factory\n",
+    "            task = task_factory(task)\n",
+    "        if task.is_classification():\n",
```
sonichi, line 880: Let's modify this notebook in a separate PR, and merge that PR at the time of new release. Otherwise, the notebook will become unusable before the new release.

markharley: Hey! Is this still the case or has the release been cut now? I can spend some time this weekend trying to get this over the line

sonichi: I mean the notebook will be unusable between the merging of this PR and the release. To prevent that, we can add a flaml version number in the beginning of the notebook's "pip install". Remove other changes for the notebook in this PR and put them in a different PR. After we release the new version which contains this PR, we merge the other PR which modifies the notebook's content as well as the version number.

markharley: Okay, sounds good. I'll get that done

markharley: Reverted now

sonichi (commented):
Please fix this error: https://github.com/microsoft/FLAML/actions/runs/4200239074/jobs/7286054080#step:12:467

markharley:
> 

Fixed now

flaml/automl/model.py
```diff
@@ -2031,7 +2040,7 @@ def search_space(cls, data_size, pred_horizon, **params):
         space.update(
             {
                 "optimize_for_horizon": {
-                    "domain": tune.choice([True, False]),
+                    "domain": tune.choice([False]),
```
sonichi, line 2043: This change modifies the original search range.

notebook/automl_classification.ipynb
```diff
@@ -867,7 +867,7 @@
     "''' SKLearnEstimator is the super class for a sklearn learner '''\n",
     "from flaml.model import SKLearnEstimator\n",
     "from flaml import tune\n",
-    "from flaml.data import CLASSIFICATION\n",
+    "from flaml.automl.task.task import CLASSIFICATION\n",
```
sonichi, line 870: In flaml version 1.1.2, this import doesn't work.

sonichi (changes requested):

flaml/automl/model.py
markharley: This is part of this change to fix a test

```
tree c6a5dd1a0e711c590b51c3e9c697eafd9fa4f7ac
parent 9df5200ddab810f72ab82800450c602c1239791a
author EgorKraevTransferwise <egor.kraev@transferwise.com> Tue Feb 7 09:58:42 2023 +0000
committer EgorKraevTransferwise <egor.kraev@transferwise.com> Tue Feb 7 09:58:42 2023 +0000

Hotwire TS_Sklearn wrapper to fix test fail by setting optimizer_for_horizon == False
```

I merged it in to get the tests passing for this PR

flaml/automl/task/task.py
```diff
@@ -0,0 +1,197 @@
...
+def get_classification_objective(num_labels: int) -> str:
+    if num_labels == 2:
+        objective_name = "binary"
+    else:
+        objective_name = "multiclass"
+    return objective_name
+
+
+class Task(ABC):
+    estimators = {}
```
coffepowered, line 62: I see this abstract class is interesting and will allow more powerful time-series tasks.
Maybe spending some words on how that's supposed to work might be useful? 

Here's my understanding:

> Abstract base class for a machine learning task. 
Class definitions should implement abstract methods and provide a non-empty dictionary of estimator classes. 
A Task can be suitable to be used for multiple machine-learning tasks (e.g. classification or regression) or be implemented specifically for a single one depending on the generality of data validation and model evaluation methods implemented. The instantiation of a Task object stores the training data and labels.
FLAML evaluates at runtime how to behave exactly, relying on `task_name`, which describes the task family.

flaml/automl/model.py
int-chaos: yea, we can disable it for now... seemed to work fine before but I'll take a look into it after this PR.

notebook/automl_xgboost.ipynb
```diff
@@ -29,7 +29,7 @@
     "\n",
     "FLAML requires `Python>=3.7`. To run this notebook example, please install flaml with the `notebook` option:\n",
     "```bash\n",
-    "pip install flaml[notebook]\n",
+    "pip install flaml[notebook]==1.0.8\n",
```
int-chaos, line 32: is there a reason why this version is 1.0.8 and not 1.1.2 since all the other ones are 1.1.2

int-chaos (commented):
everything else looks good to me!

notebook/automl_xgboost.ipynb
markharley: Good point, not sure why. I'll bump it. Thanks!

flaml/automl/task/task.py
markharley: Thanks! It's a good point and very helpful of you to include a docstring for me :D. I'll make that change today and add docstrings throughout the new ABC

coffepowered: Great! I am happy I got the main intent and thank you for your response.

Still, I had not a full understanding of the following points when I wrote that:
1. why X,y are necessary at the initialization,
2. the fact that the `Task` class is a different concept wrt the "machine learning term to designate a family of problems" (full awareness came later).

might be the opportunity to improve my docstring draft a bit :)

flaml/automl/model.py
sonichi: > yea, we can disable it for now... seemed to work fine before but I'll take a look into it after this PR.

Allowing this change here would decrease the code coverage and make buggy code stay in the repo without being covered by test. The right solution is to either remove or fix the buggy part. For example, if for some reason we don't support `optimize_for_horizon`=True, we should either fix it or remove the code corresponding to that case, and document the reason it doesn't work.

notebook/automl_xgboost.ipynb
markharley: I've bumped it now and pushed

flaml/automl/task/task.py
markharley: Thanks for the feedback. I've used your docstring as a start and modified to hopefully answer the additional questions listed

flaml/automl/ml.py
```diff
@@ -674,14 +561,13 @@ def compute_estimator(
             free_mem_ratio=0,
         )
     else:
-        val_loss, metric_for_logging, train_time, pred_time = evaluate_model_CV(
+        val_loss, metric_for_logging, train_time, pred_time = task.evaluate_model_CV(
```
sonichi, line 564: In future, `get_val_loss` can be moved into Task too.

flaml/automl/task/generic_task.py
```diff
@@ -0,0 +1,849 @@
+import logging
```
sonichi, line 1: Most code in this file is moved from automl.py. I didn't verify whether there is any unnecessary change when moving. I'll appreciate if another reviewer could check it.

flaml/automl/model.py
```diff
@@ -1486,8 +1493,12 @@ def cost_relative2lgbm(cls):
         return 1.9
 
     def __init__(self, task="binary", **params):
+        if isinstance(task, str):
+            from flaml.automl.task.factory import task_factory
+
+            task = task_factory(task)
         super().__init__(task, **params)
-        if "regression" in task:
+        if task.is_regression():
```
sonichi, line 1501: There are more issues like this: 1453, 1470, 1484. There should be more.

flaml/automl/model.py
```diff
@@ -2122,11 +2133,7 @@ def predict(self, X, **kwargs):
                         X.iloc[:i, :]
                     )
                     preds.append(self._model[i - 1].predict(X_pred, **kwargs)[-1])
-                forecast = DataFrame(
-                    data=np.asarray(preds).reshape(-1, 1),
-                    columns=[self.hcrystaball_model.name],
-                    index=X.index,
-                )
+                forecast = Series(preds)
```
sonichi, line 2136: @int-chaos please check this.

int-chaos: Yes I did this. this fixes the issue with automl.py/predict(), where converting the DataFrame into a Series was causing an error. There's no fundamental change to the overall output since in automl.py/predict() when self._label_transformer is true, it converts it to a pd.Series.

flaml/automl/model.py
int-chaos: why is this an issue?

EgorKraevTransferwise: `task in [list]` is actually perfectly legitimate if `task` is a `Task` object, as the latter implements a comparison operator to strings. There are so many of these scattered all over the code that we felt it was easier to support them as they are for now, and gradually clean them up later.

flaml/automl/model.py
EgorKraevTransferwise: We could change the test prediction horizon to have the same length as the validation horizon? :)
The long-term fix would be to predict recursively, that is already done in the follow-up time series PR

flaml/automl/model.py
sonichi: > Yes I did this. this fixes the issue with automl.py/predict(), where converting the DataFrame into a Series was causing an error. There's no fundamental change to the overall output since in automl.py/predict() when self._label_transformer is true, it converts it to a pd.Series.

Could you create an issue to update the notebooks and examples in the documentation website?

flaml/automl/model.py
EgorKraevTransferwise: On the other hand, `"regression" in task` would not work with that approach; but I did a full-text search of the code for `in task` and found no hits

flaml/automl/automl.py
```diff
@@ -2539,7 +1535,10 @@ def cv_score_agg_func(val_loss_folds, log_metrics_folds):
 
         self._state._start_time_flag = self._start_time_flag = time.time()
         task = task or self._settings.get("task")
-        self._estimator_type = "classifier" if task in CLASSIFICATION else "regressor"
+        if isinstance(task, str):
```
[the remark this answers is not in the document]
EgorKraevTransferwise: Done in all of automl.py

flaml/automl/ml.py
```diff
@@ -354,8 +356,8 @@ def sklearn_metric_loss_score(
     return score
 
 
-def get_y_pred(estimator, X, eval_metric, obj):
-    if eval_metric in ["roc_auc", "ap", "roc_auc_weighted"] and "binary" in obj:
+def get_y_pred(estimator, X, eval_metric, task):
```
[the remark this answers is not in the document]
EgorKraevTransferwise: Done

flaml/default/suggest.py
```diff
@@ -45,6 +46,7 @@ def meta_feature(task, X_train, y_train, meta_feature_names):
 
 
 def load_config_predictor(estimator_name, task, location=None):
+    task = str(task)
```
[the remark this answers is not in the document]
EgorKraevTransferwise: Not sure I understand. This will accept either str- or Task-valued inputs, what's there to check?

flaml/automl/automl.py
```diff
@@ -1661,10 +680,7 @@ def retrain_from_log(
         Args:
             log_file_name: A string of the log file name.
             X_train: A numpy array or dataframe of training data in shape n*m.
-                For time series forecast tasks, the first column of X_train
-                must be the timestamp column (datetime type). Other
-                columns in the dataframe are assumed to be exogenous
-                variables (categorical or numeric).
+                For time series forecast tasks, the first column of X_train must be the timestamp column (datetime type). Other columns in the dataframe are assumed to be exogenous variables (categorical or numeric).
```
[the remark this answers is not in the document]
EgorKraevTransferwise: CI formatting check passes, what specific issues should we check?

flaml/automl/task/task.py
```diff
@@ -0,0 +1,354 @@
...
+                such as in binary vs multilabel classification.
+        """
+        self.name = task_name
+
+    def __str__(self) -> str:
+        """Name of this task type."""
+        return self.name
+
+    @abstractmethod
+    def evaluate_model_CV(
```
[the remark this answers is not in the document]
EgorKraevTransferwise: As far as I know, there's no need to modify `get_val_loss` according to task type, so why should it be moved to here? Could we say we leave it for now, and move it to the Task class if there ever arises a need to for different implementations according to task?

qingyun-wu (approved):
Thanks for addressing the comments. The PR looks good to me now.

sonichi (approved):
Please address https://github.com/microsoft/FLAML/pull/857#discussion_r1129592420

I'll merge after that.

Issue #866, opened 2022-12-28
Low cpu usage when parallel tuning


shufan007:
Hi, first of all thanks for Your kindly reply and support.
I got a low cpu uasge wen doing parallel tuning with blendsearch. My code and parameters as below:

```
from flaml import AutoML
automl = AutoML()
automl.fit(X_train, y_train, 
        task="classification",
        estimator_list=["lgbm"],
        metric='roc_auc',
        time_budget=600, 
        n_jobs=8,
        n_concurrent_trials=2,
        hpo_method='bs')
```
Here , X_train is a ndarray with size [20000000, 39] and dtype 'float64' ,  the y_train's  dtype is 'int32'.
The task is executed in a container with 16CPU, 180G memory resources

This got a very low cpu uasge, only around 200%, not the expected 1600%. So the training trails are vary slow.

Supplementary Note：

During trails running, the following warning occurred:
WARNING trial_runner.py:886 -- Trial Runner checkpointing failed: cannot serialize a bytes object larger than 4 GiB

when I cut the size of X_train to [10000000, 39], everything looks normal, the cpu uasge reached close to 1600%.

It seems to be related to the data size of X_train, for
20000000 x 39 x 8/(1024^3) \approx 5.8GiB > 4 GiB
10000000 x 39 x 8/(1024^3) \approx 2.9GiB < 4 GiB 

Please could You kindly help me, Thanks!

sonichi:
@Yard1 This seems an issue caused by ray. Do you have an idea?

Yard1:
Can you point me to where you are specifying how to pass the data to Tune? @sonichi

sonichi:
> Can you point me to where you are specifying how to pass the data to Tune? @sonichi
> 
> It appears that the data is serialized in the checkpoint, which shouldn't happen unless it's passed as a part of the config, or as a part of the reported metrics. Using `with_parameters` should not cause this behavior. This may also be caused if the searcher, scheduler or trial runner have the data as an attribute, as the entire object will be pickled on checkpoint. I'd imagine the low CPU utilization is due to serialization overheads of the data being passed, which causes the trial runner loop to get bottlenecked.

Here it is: https://github.com/microsoft/FLAML/blob/b78d0b57c498ab852aec7f8d30bc782d7914ff69/flaml/automl/automl.py#L2144

Yard1:
@shufan007 It it possible for you to provide a fully self contained reproducible example?

shufan007:
[...]

PR #870, opened 2023-01-04
Do not persist entire AutoMLState in Searcher

Yard1:
Signed-off-by: Antoni Baum <antoni.baum@protonmail.com>

Currently, the entire `AutoMLState` is persisted in Searcher as a part of the `config_constraints`. As the State object contains the dataframe used, this means a large amount of data will be persisted to disk as a part of automatic Tune Experiment checkpointing, leading to large disk usage and high performance overhead, causing various bottlenecks.

This PR changes the `size` function used in `config_constraints` to only consider the `learner_classes` attribute of `AutoMLState` which removes the issue of too much data being saved.

## Related issue number

Closes https://github.com/microsoft/FLAML/issues/866

## Checks

### DIFF
--- a/flaml/automl/automl.py
+++ b/flaml/automl/automl.py
@@ -491,7 +491,7 @@ class AutoMLState:
         return estimator, train_time
 
 
-def size(state: AutoMLState, config: dict) -> float:
+def size(learner_classes: dict, config: dict) -> float:
     """Size function.
 
     Returns:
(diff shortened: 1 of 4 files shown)


sonichi (approved):
Looks good to me. Thanks!

flaml/automl/automl.py
```diff
@@ -491,15 +491,15 @@ def _train_with_config(
         return estimator, train_time
 
 
-def size(state: AutoMLState, config: dict) -> float:
+def size(learner_classes: dict, config: dict) -> float:
```
thinkall, line 494: Is this function actually used? I don't find any call of it.

thinkall: My bad. Thanks.

thinkall (approved):

Issue #771, opened 2022-10-25
pytorch TimeSeriesDataSet handling missing time stamps - for TemporalFusionTransformer model


int-chaos:
PyTorch's TimeSeriesDataSet is capable of handling missing time stamps. We would like to leverage this capability for our TemporalFusionTransformer model. PyTorch has two methods forward fill strategy and constant fill strategy. 
- Set `allow_missing_timesteps=True`
- Add fill strategies to search space when there exist missing data. 
- For constant fill strategies, will need to search for constant values to use for each column name.

See [PyTorch TimeSeriesDataSet Documentation](https://pytorch-forecasting.readthedocs.io/en/stable/api/pytorch_forecasting.data.timeseries.TimeSeriesDataSet.html) for more information.

int-chaos:
This would also affect our way of dealing with missing timestamps.

Current assumption is that users do no have missing timestamps in their time series and we use `pd.infer_freq()` and `pd.data_range` to help create our `time_idx` column and output an error message if `pd.infer_freq()` is `None` (missing timestamps).

With this new update, users have two options:
 1. input a timeseries with no missing timestamps and we continue using `pd.infer_freq()` and `pd.data_range` to help create our `time_idx` column
 2. input a `freq` argument. this way we can still use `pd.data_range` to help create our `time_idx` column

PR #878, opened 2023-01-09
allow missing data for "ts_forecast_panel" task

### LINKED ISSUE #754: Time series gap detection for TFT tasks

int-chaos:
- allow missing data for "ts_forecast_panel" task
- address typo for `hcrystalball_model`
- add test for missing data scenario in test_forecast.py

## Related issue number
- closes #754 #770 
- addresses part 1 of #771 
- closes #839

### DIFF
--- a/flaml/automl/automl.py
+++ b/flaml/automl/automl.py
@@ -1054,12 +1054,14 @@ class AutoML(BaseEstimator):
             assert (
                 dataframe[[dataframe.columns[0]]].duplicated() is None
             ), "Duplicate timestamp values with different values for other columns."
-        ts_series = pd.to_datetime(dataframe[dataframe.columns[0]])
-        inferred_freq = pd.infer_freq(ts_series)
-        if inferred_freq is None:
-            logger.warning(
-                "Missing timestamps detected. To avoid error with estimators, set estimator list to ['prophet']. "
-            )
+        if self._state.task in TS_FORECAST and not TS_FORECASTPANEL:
+            # TFT estimator model for TS_FORECASTPANEL can handle missing data
+            ts_series = pd.to_datetime(dataframe[dataframe.columns[0]])
+            inferred_freq = pd.infer_freq(ts_series)
+            if inferred_freq is None:
+                logger.warning(
+                    "Missing timestamps detected. To avoid error with estimators, set estimator list to ['prophet']. "
+                )
         if y_train_all is not None:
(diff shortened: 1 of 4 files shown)


int-chaos:
there is a current error with missing timestamps test. please advise if you can.

missing timestamps causes error with cross validation. 

With the current seeding, 04/01/2017 data is removed for Agency_13  SKU_04, but the time range 01/01/2017 - 06/01/2017 is used as the testing portion for holdout validation. Even though 04/01/2017 Agency_13 SKU_04 is not part of the testing data (in this case testing data is 01/01/2017 - 06/01/2017 for all agency and sku groups, with the exception of 04/01/2017 Agency_13 SKU_04), when the model predicts it still predicts a value for 04/01/2017 Agency_13 SKU_04, thus resulting in a ValueError where the testing length and prediction length are different. 

I've been trying to come up with a way to remove the prediction for 04/01/2017 Agency_13 SKU_04, but there is no easy way to locate it since the raw prediction is just a tensor.
image

sonichi:
> there is a current error with missing timestamps test. please advise if you can.
> 
> missing timestamps causes error with cross validation.
> 
> With the current seeding, 04/01/2017 data is removed for Agency_13 SKU_04, but the time range 01/01/2017 - 06/01/2017 is used as the testing portion for holdout validation. Even though 04/01/2017 Agency_13 SKU_04 is not part of the testing data (in this case testing data is 01/01/2017 - 06/01/2017 for all agency and sku groups, with the exception of 04/01/2017 Agency_13 SKU_04), when the model predicts it still predicts a value for 04/01/2017 Agency_13 SKU_04, thus resulting in a ValueError where the testing length and prediction length are different.
> 
> I've been trying to come up with a way to remove the prediction for 04/01/2017 Agency_13 SKU_04, but there is no easy way to locate it since the raw prediction is just a tensor. image

What does it mean by `With the current seeding, 04/01/2017 data is removed for Agency_13 SKU_04"? Why is it removed?

PR #888, opened 2023-01-22
add cost budget; move loc of make_dir

sonichi:

1. Allow users to specify a cost budget when cost_attr is neither time nor num_samples in BlendSearch.
2. create a directory for log_file_name even when use_ray is true, to avoid FileNotFoundError.

## Checks


flaml/tune/searcher/blendsearch.py
```diff
@@ -693,6 +700,8 @@ def on_trial_result(self, trial_id: str, result: Dict):
     def suggest(self, trial_id: str) -> Optional[Dict]:
         """choose thread, suggest a valid config."""
         if self._init_used and not self._points_to_evaluate:
+            if self._cost_budget and self._cost_used >= self._cost_budget:
+                return None
```
ZviBaratz, line 704:
```suggestion
                return 
```

flaml/tune/searcher/blendsearch.py
```diff
@@ -154,8 +157,10 @@ def __init__(
                 self.cost_attr = TIME_TOTAL_S
             else:
                 self.cost_attr = None
+            self._cost_budget = None
         else:
             self.cost_attr = cost_attr
+            self._cost_budget = cost_budget
```
liususan091219, line 163: why is there _?

Yard1 (approved):
Looks fine to me from the Ray side

flaml/tune/searcher/blendsearch.py
sonichi: internal variable.

flaml/tune/searcher/blendsearch.py
thinkall: Why remove "None" from "return None"? `suggest` should return `config` if it ends sucessfully, I think `return None` here indicates that this function should return some value, just at this point, it returns None.

Considering `def suggest(self, trial_id: str) -> Optional[Dict]:`, maybe we can use `return {}`?

qingyun-wu (approved):

Issue #917, opened 2023-02-10
[nlp] KeyError when loading the bert model


liususan091219:
When running the notebook with the bert model, the following error shows up. It is because the transformers version in setup.py is outdated. Once updated to latest version of transformers, this problem is gone. 

```python
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
[<ipython-input-35-382ffd541e62>](https://localhost:8080/#) in <module>
     39 }
     40 
---> 41 automl_model.fit(X_train=X_train, y_train=y_train,X_val=X_val, y_val=y_val, custom_hp=custom_hp, **automl_settings)

14 frames
[/usr/local/lib/python3.8/dist-packages/transformers/modeling_utils.py](https://localhost:8080/#) in _load_pretrained_model(cls, model, state_dict, resolved_archive_file, pretarained_model_name_or_path, ignore_mismatched_sizes, sharded_metadata, _fast_init)
   2000                     if (
   2001                         model_key in model_state_dict
-> 2002                         and state_dict[checkpoint_key].shape != model_state_dict[model_key].shape
   2003                     ):
   2004                         mismatched_keys.append(

KeyError: 'bert.embeddings.LayerNorm.weight'
```

PR #918, opened 2023-02-10
Update hf version

liususan091219:

To fix the issue in Issue #917 , i.e., the hf version is outdated that causes the bert model to have key error. 

## Related issue number

#917 

## Checks

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,7 +63,7 @@ setuptools.setup(
             "statsmodels>=0.12.2",
             "psutil==5.8.0",
             "dataclasses",
-            "transformers[torch]==4.18",
+            "transformers[torch]",
             "datasets",
             "nltk",
             "rouge_score",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -92,7 +92,7 @@
             "vowpalwabbit>=8.10.0, <9.0.0",
         ],
         "nlp": [
-            "transformers[torch]==4.18",
+            "transformers[torch]",
```
sonichi, line 95: What about using a specific version here? That way, we can make sure when the user installs the [nlp] option, the code doesn't stop working when future version of transformers is released.

setup.py
```diff
@@ -63,7 +63,7 @@
             "statsmodels>=0.12.2",
             "psutil==5.8.0",
             "dataclasses",
-            "transformers[torch]==4.18",
+            "transformers[torch]",
```
sonichi, line 66: This change is OK as it would test the latest version of transformers in the build action, so that we would know when a future release of transformers breaks our test.

sonichi:
Please fix the typo in the title.

setup.py
liususan091219: Updated

sonichi (approved):

ZviBaratz (approved):

qingyun-wu (approved):

PR #926, opened 2023-02-22
handling nlp divide by zero

liususan091219:

NLP module has ZeroDivisionError: https://github.com/microsoft/FLAML/actions/runs/4220014982/jobs/7325940952#step:12:448 which happens for:

```python
automl.predict(["test test", "test test"])
```
This error happens because, in transformers, one variable is divided by the runtime of this line: https://github.com/huggingface/transformers/blob/main/src/transformers/trainer_utils.py#L351. 

This PR catches this exception and pass. 

## Checks


flaml/automl/automl.py
```diff
@@ -994,7 +997,10 @@ def predict(
             )
             return None
         X = self._preprocess(X)
-        y_pred = estimator.predict(X, **pred_kwargs)
+        try:
+            y_pred = estimator.predict(X, **pred_kwargs)
+        except ZeroDivisionError:
+            y_pred = 0
```
sonichi, line 1003: I don't think this is the right place to catch the error because the error is specific to transformers.

test/nlp/test_autohf.py
```diff
@@ -62,7 +62,12 @@ def test_hf_data():
         **automl_settings
     )
     automl.predict(X_test, **{"per_device_eval_batch_size": 2})
-    automl.predict(["test test", "test test"])
+    automl.predict(
```
sonichi, line 65: Please add a test to make sure the ZeroDivisionError is raised. For example, using an even shorter text. Please verify that ZeroDivisionError is raised and your handling of the error is correct.

thinkall:
`FAILED test/nlp/test_autohf.py::test_hf_data - KeyError: "['sentence2'] not in index"`
https://github.com/microsoft/FLAML/actions/runs/4249519669/jobs/7389744926#step:12:406

sonichi:
https://github.com/microsoft/FLAML/actions/runs/4253860408/jobs/7399306260#step:12:196

website/docs/Examples/Integrate - OpenAI.md
```diff
@@ -0,0 +1,158 @@
...
+
+Setup your OpenAI key:
+```python
+import os
+
+if "OPENAI_API_KEY" not in os.environ:
+    os.environ["OPENAI_API_KEY"] = "<your OpenAI API key here>"
+```
+
+### Start an
```
sonichi, line 17: Incomplete

website/docs/Examples/Integrate - OpenAI.md
```diff
@@ -0,0 +1,158 @@
+FLAML has integrated the OpenAI's completion API.
```
sonichi, line 1: This will be the first sentence a reader of the paper will see. It could be first time they visit FLAML. Think of what to say here to keep them interested in reading the rest.

website/docs/Examples/Integrate - OpenAI.md
```diff
@@ -0,0 +1,158 @@
...
+
+#### Run distributed tune job
+
+Prepare a script like [ray/distribute_tune.py](https://github.com/microsoft/FLAML/blob/main/test/ray/distribute_tune.py). Replace the command in the above eample with:
+
+```python
+command = ["python distribute_tune.py"]
+```
+
+Everything else is the same.
```
sonichi, line 158: These are not about OpenAI

flaml/automl/model.py
```diff
@@ -837,8 +837,11 @@ def predict_proba(self, X, **pred_kwargs):
         test_dataset = Dataset.from_pandas(X_test)
 
         new_trainer = self._init_model_for_predict()
-        predictions = new_trainer.predict(test_dataset)
-        return predictions.predictions
+        try:
+            predictions = new_trainer.predict(test_dataset)
+            return predictions.predictions
+        except ZeroDivisionError:
+            return []
```
sonichi, line 844: Can the consumer of predictions (i.e., the evaluator) handle the empty list as returned value? I doubt it.

liususan091219: previously I tried to return None and now with the return []. If neither works, do you have any recommendations on how to handle the ZeroDivisionError? Maybe print a warning and return None?

sonichi (approved):

flaml/automl/model.py
```diff
@@ -1191,8 +1191,12 @@ def predict_proba(self, X, **pred_kwargs):
...
 
         new_trainer = self._init_model_for_predict()
-        predictions = new_trainer.predict(test_dataset)
-        return predictions.predictions
+        try:
+            predictions = new_trainer.predict(test_dataset)
+            return predictions.predictions
+        except ZeroDivisionError:
+            logger.warning("Zero division error appeared in HuggingFace Transformers.")
+            return np.array([-0.05] * len(test_dataset))
```
ZviBaratz, line 1199:
```suggestion
        try:
            predictions = new_trainer.predict(test_dataset)
        except ZeroDivisionError:
            logger.warning("Zero division error appeared in HuggingFace Transformers.")
            return np.array([-0.05] * len(test_dataset))
        else:
            return predictions.predictions       
```

flaml/automl/model.py
```diff
@@ -1222,21 +1226,26 @@ def predict(self, X, **pred_kwargs):
...
-            y_pred=predictions.predictions,
-            tokenizer=self.tokenizer,
-            hf_args=self._training_args,
-            X=X,
-        )
-        return post_y_pred
+            return post_y_pred
+        except ZeroDivisionError:
+            logger.warning("Zero division error appeared in HuggingFace Transformers.")
+            return np.array([0] * len(test_dataset))
```
ZviBaratz, line 1248: Isn't it just the `predict()` calls that could raise the exception? Perhaps it would be better to create a `kwargs` variable that will be either empty (`{}`) or `{"metric_key_prefix": "predict"}`, and then call once with a shorter `try-except` clause? 
AFAIK, it is not good practice to include this much code within a `try-except` clause.

ZviBaratz (commented):
Made a couple of suggestions. It looks OK, I just think the contents of `try-except` clauses should be as specific as possible.

flaml/automl/model.py
liususan091219: Thanks for the comments; addressed.

ZviBaratz (approved):

PR #928, opened 2023-02-22
minor change to notebook

sonichi:

Fix an indent and version number in the openai notebook.

## Checks


thinkall (commented):
Looks good to me. Just wondering why do we need to fix the version?

qingyun-wu (approved):

sonichi:
> Looks good to me. Just wondering why do we need to fix the version?

Version 1.1.3 updated the price from v1.1.2. The notebook's results will be quite different when using these two different versions, including how long it takes to finish.

thinkall:
> > Looks good to me. Just wondering why do we need to fix the version?
> 
> Version 1.1.3 updated the price from v1.1.2. The notebook's results will be quite different when using these two different versions, including how long it takes to finish.

`pip install` will always install the latest version by default. To avoid installing the eariler version, maybe better use `>=1.1.3`.

sonichi:
> > > Looks good to me. Just wondering why do we need to fix the version?
> > 
> > 
> > Version 1.1.3 updated the price from v1.1.2. The notebook's results will be quite different when using these two different versions, including how long it takes to finish.
> 
> `pip install` will always install the latest version by default. To avoid installing the eariler version, maybe better use `>=1.1.3`.

After we release a new version and before we update the notebook, if a user executes `pip install flaml>=1.1.3`, they will see an inconsistent result. That's why I fixed the version number here.

thinkall (approved):

PR #931, opened 2023-02-23
ci: concurrency

Borda:

Adding concurrency to the most time consuming builds so only the last commit within PR is build and canceled ll previous, which would significantly lower the action utilization all fet results faster... Also this is only for PR so any commit to main branch will be run always

## Checks


.github/workflows/python-package.yml
```diff
@@ -17,11 +17,16 @@ on:
   merge_group:
     types: [checks_requested]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.head_ref }}
+  cancel-in-progress: ${{ ! (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/heads/release/')) }}
```
sonichi, line 22: I don't know the meaning of `github.ref` and whether 'refs/heads/master' is a valid value. Our main branch is "main" in case they are related.

.github/workflows/python-package.yml
```diff
@@ -17,11 +17,16 @@ on:
...
+
 jobs:
   build:
 
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
```
sonichi, line 29: What's the reason for this line?

sonichi (commented):
Thank you. This feature would be very useful, although I don't know the correctness of the implementation. I hope other reviewers who understand the implementation could help check it.

.github/workflows/python-package.yml
Borda: it was quite confusing to see jobs canceled anot be able to know if all is fine except one

Borda:
> Thank you. This feature would be very useful, although I don't know the correctness of the implementation. I hope other reviewers who understand the implementation could help check it.

you can check the official docs: https://docs.github.com/en/actions/examples/using-concurrency-expressions-and-a-test-matrix

Borda:
helo, @thinkall, @sonichi, @markharley, @ZviBaratz, @coffepowered could you pls have look? :chipmunk:

ZviBaratz:
I don't have experience with concurrency in GitHub actions, but I tried looking at the docs and as far as I can tell, it looks good. Could you please just clarify the purpose of including `${{ github.head_ref }}` in the group? I'm not sure what would happen in non PR events (where this property is undefined).

.github/workflows/python-package.yml
```diff
@@ -17,11 +17,16 @@ on:
   merge_group:
     types: [checks_requested]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.head_ref }}
```
thinkall, line 21: Looks like it's possible to trigger a syntax error if not in a PR workflow. 
https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-a-fallback-value

And the current workflow does server more than pull_request.

```

on:
  push:
    branches: ['main']
    paths:
      - 'flaml/**'
      - 'test/**'
      - 'notebook/**'
      - '.github/workflows/python-package.yml'
      - 'setup.py'
  pull_request:
    branches: ['main']
  merge_group:
    types: [checks_requested]
```
```suggestion
  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.head_ref || github.run_id }}
```

.github/workflows/python-package.yml
```diff
@@ -17,11 +17,16 @@ on:
...
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
```
thinkall, line 22: Why not just make it `true`. It looks like you just want to cancel jobs not in main branch, but this workflow will only run on main branch. Plus, make cancel to be `true` will only cancel jobs in the same group, i.e., the same PR or same push to main.

thinkall (changes requested):

.github/workflows/python-package.yml
Borda: > make cancel to be `true` will only cancel jobs in the same group, i.e., the same PR or same push to main

exactly, bu you do not want to cancel any in you default/main branch

.github/workflows/python-package.yml
Borda: > And the current workflow does server more than pull_request.

Not sure if I understand what you mean...

thinkall: The workflow can be triggered by push to main, in that case it will raise error

.github/workflows/python-package.yml
thinkall: Do you mean cancel the runs in your forked repo? I don't think it's needed to trigger runs in the forked main branch. I know the current workflow will trigger it, but it doesn't matter. Please correct me.

Borda: no, all this CI is running on your side and by group, you define a pipe where workflows are executed sequentially. The second `cancel-in-progress` stated that all such pipelines (except the `main`) execute only the last workflow and cancel all previous ones even they are still running

.github/workflows/python-package.yml
Borda: what kind of error? if a variable is not set GH gives empty string not error

thinkall: I didn't verify it. Just read it from the doc https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-a-fallback-value

.github/workflows/python-package.yml
thinkall: image

I believe the CI is running on FLAML side. 

For the cancel param, in the [doc](https://docs.github.com/en/actions/using-jobs/using-concurrency#overview), it says: 
```
To also cancel any currently running job or workflow in the same concurrency group, specify cancel-in-progress: true
```
As far as I understand, it's exactly what we want with concurrency. Would you mind post some figures to show your point. I'm a little bit confused. Sorry for raising the questions and thank you so much for your patience.

Borda: > As far as I understand, it's exactly what we want with concurrency. Would you mind post some figures to show your point. I'm a little bit confused. Sorry for raising the questions and thank you so much for your patience.

Yes but in case you have several fast merged PRs yielding in commits to main brach you want let finish CI for each of them not just the last one...

.github/workflows/python-package.yml
Borda: telling from experience as I have implemented this on other projects :chipmunk:

.github/workflows/python-package.yml
thinkall: > > As far as I understand, it's exactly what we want with concurrency. Would you mind post some figures to show your point. I'm a little bit confused. Sorry for raising the questions and thank you so much for your patience.
> 
> Yes but in case you have several fast merged PRs yielding in commits to main brach you want let finish CI for each of them not just the last one...

So if the group name is `group: github.workflow−{{ github.ref }}-${{ github.head_ref || github.run_id }}`, we can safely set true here. Right?

Borda: if you want to procced only with the last commit to master/main branch, then yes

Borda: So let's have a model situation CI that takes one hour, and you have pushed 5 times a commit every 10 minutes (times are just to illustrate the case, not realistic atm)

- `cancel-in-progress: true`
  - for branch or PR, when a new commit is pushed, that running CI is stopped
- `cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}`
  - for PR, when a new commit is pushed, that running CI is stopped
  - for the main branch always waits to finish CI

so for open PR the CI will take at most 1hour as all previous jobs are canceled
but for main, you will have finished CI always check so when you check the commit history, you would always know that a particular commit was valid

thinkall (approved):

ZviBaratz (approved):

ZviBaratz (approved):

PR #948, opened 2023-03-12
Fix typo in integrate_chatgpt_code.ipynb

eltociear:
overriden -> overridden

PR #958, opened 2023-03-20
gpt-4 support; openai workflow fix; model str; timeout; voting

sonichi:

1. Notebook file added to workflow paths. 
2. Allow model of str type in `oai.Completions.tune()`.
3. Catch the Timeout Error from openai.
4. Notebook typo fix. Credit to @eltociear 's #948 . The OpenAI workflow can't be executed in a forked repo, so #948 can't be merged.
5. Use voting for the math benchmark.
6. Support gpt-4.

## Checks


sonichi:
> https://github.com/microsoft/FLAML/blob/03f2612fee09502e9a0848367ae7005f9aa46b89/flaml/integrations/oai/completion.py#L138-L140
> 
> Is "RateLimitError" still a proper phrase here?

Yes, because we only increase `retry` for RateLimitError.

sonichi:
> with timeout_limit(timeout=30):

`timeout_limit` is not a native function. What package do you suggest to install?

notebook/integrate_chatgpt_math.ipynb
kevin666aa, line 1: It takes $0.2 to run the default config but $0.87 to run the searched config on test. Is it fare to compare them?

sonichi: How about adding a note that the default config has lower inference budget? And a note that flaml allows developers to tune the config according to their target inference budget?

thinkall:
> > with timeout_limit(timeout=30):
> 
> `timeout_limit` is not a native function. What package do you suggest to install?

Would the one you implemented in `model.py` work? The `limit_resource`.

sonichi:
> > > with timeout_limit(timeout=30):
> > 
> > 
> > `timeout_limit` is not a native function. What package do you suggest to install?
> 
> Would the one you implemented in `model.py` work? The `limit_resource`.

There appears to be a timeout argument in openai api. I'll test that out.

sonichi:
> > > > with timeout_limit(timeout=30):
> > > 
> > > 
> > > `timeout_limit` is not a native function. What package do you suggest to install?
> > 
> > 
> > Would the one you implemented in `model.py` work? The `limit_resource`.
> 
> There appears to be a timeout argument in openai api. I'll test that out.

It doesn't make an effect now. But it make work in future. If it works, then users can simply pass timeout in `tune()`. How about revisiting this issue in future?

notebook/integrate_chatgpt_math.ipynb
kevin666aa: I think the main problem is that the tuned config get 9 responses and choose the answer with the most voting, but the default config gets only one response. I think we should allow more responses for the default config and also voting so that it uses approximate the same or even bigger inference budget.

sonichi: I added gpt-4 support. After adding it, it's not obvious what I should change about the default config. The default config now uses gpt-4 with n=1. The tuned config uses gpt-3.5-turbo with n=53.

kevin666aa (approved):

qingyun-wu (approved):

PR #975, opened 2023-04-04
set black with 120 line length

Borda:
## Why are these changes needed?

as Flake8 is set already to 127 line length Black has default 88 so it is natural to have both aligned, see flake8 setting: https://github.com/microsoft/FLAML/blob/f404b66a4c20d421c9cee465a6796ca76a0a8992/.flake8#L3

## Checks

- [x] I've used [pre-commit](https://microsoft.github.io/FLAML/docs/Contribute#pre-commit) to lint the changes in this PR, or I've made sure [lint with flake8](https://github.com/microsoft/FLAML/blob/816a82a1155b4de4705b21a615ccdff67c6da379/.github/workflows/python-package.yml#L54-L59) output is two 0s.
- [ ] I've included any doc changes needed for https://microsoft.github.io/FLAML/. See https://microsoft.github.io/FLAML/docs/Contribute#documentation to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [x] I've made sure all auto checks have passed.

PR #976, opened 2023-04-04
pyproject.toml & switch to Ruff

Borda:
### unify config to `pyproject.toml`

as of now, there are configs for each tool which make discoverability harder (just a bit), so unification will also facilitate and easily spot uncoherent cross-tool setting like line-length

### replace flake8 with Ruff

It is quite a new tool that has the same issue coverage as `flake8,` but it is 10-100x faster, which is quite convenient for the linter. Moreover, it also besides raising issue perform some simple fixing

## Related issue number

uses already set Blak line length from #975

## Checks


Borda:
@sonichi @thinkall @ZviBaratz could you pls have look? :rabbit:

pyproject.toml
```diff
@@ -0,0 +1,51 @@
...
+]
+
+[tool.black]
+# https://github.com/psf/black
+line-length = 120
+exclude = "(.eggs|.git|.hg|.mypy_cache|.venv|_build|buck-out|build|dist)"
+
+
+[tool.ruff]
+line-length = 127
```
ZviBaratz, line 19: Shouldn't this match `black`'s config?
```suggestion
line-length = 120
```

Borda: it should; I just took the same config as it was for flake8

ZviBaratz (approved):

thinkall:
Thank you for the improvements! @Borda 
I totally agree with the pyproject.toml part. Regarding to Ruff, I'm not sure if it's necessary. Ruff is only at v0.0.261 and flake8 is not too slow for flaml.

WDYT, @sonichi ?

Borda:
> I'm not sure if it's necessary. Ruff is only at v0.0.261 and flake8 is not too slow for flaml.

The main advantage for me is that `ruff` also performs some simple fixes compare to `flake8`
and regarding the number, it has been adopted by several major projects like Scipy, Pandas, FastAPI

sonichi:
> Thank you for the improvements! @Borda I totally agree with the pyproject.toml part. Regarding to Ruff, I'm not sure if it's necessary. Ruff is only at v0.0.261 and flake8 is not too slow for flaml.
> 
> WDYT, @sonichi ?

No strong opinion. Happy to follow the community's preference.

thinkall (approved):

pyproject.toml
sonichi, line 1: What will be the effect of adding this file? Would it affect the pypi packaging in next release?

sonichi: Thanks. Could you explain how this file is used? I guess the [tool.ruff] section is going to be used by ruff.

Borda: Correct, pyproject.toml is standard python config file https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/ or https://stackoverflow.com/a/66472800/4521646

sonichi (changes requested):
Please modify the PULL_REQUEST_TEMPLATE about the flake8 part.

pyproject.toml
```diff
@@ -0,0 +1,51 @@
+[metadata]
+license_file = "LICENSE"
+description-file = "README.md"
+
+
+[tool.pytest.ini_options]
+addopts = '-m "not conda"'
+markers = [
+    "conda: test related to conda forge distribution"
+]
```
sonichi, line 10: @MichalChromcak @skzhang1 could you help review this part as it seems related to conda?

sonichi (approved):
I suggest merging after @MichalChromcak or @skzhang1 confirms that the change doesn't affect the conda packaging.

MichalChromcak:
> I suggest merging after @MichalChromcak or @skzhang1 confirms that the change doesn't affect the conda packaging.

I think, it should be fine. What to watch out in general is/are:
- package versions (some functionality might not work on old ones)
- coverage of the functionality of the tests marked with conda (are we really testing enough functionality in the conda build ci?)

In ancy case, updating [this recipe](https://github.com/conda-forge/flaml-feedstock/blob/main/recipe/meta.yaml) should be enough.

MichalChromcak (approved):

Borda:
> In any case, updating [this recipe](https://github.com/conda-forge/flaml-feedstock/blob/main/recipe/meta.yaml) should be enough.

Changes in this PR does not require changes in conda receipt (by my best knowledge)

PR #989, opened 2023-04-10
Factor out time series-related functionality into a time series Task object

EgorKraevTransferwise:

This continues the process that began with the task-based-refactor PR, now merged, Time-series related models are moved into a subpackage, related bits in the Task object are moved into a separate Task. In the process, some existing bugs are fixed (such as hcrystalball wrapper with certain configs only supporting forecasting for as far ahead as the length of the validation set).
Also, input data is auto-enriched with features derived from the timestamps.
## Checks

- [ ] I've used [pre-commit](https://microsoft.github.io/FLAML/docs/Contribute#pre-commit) to lint the changes in this PR, or I've made sure [lint with flake8](https://github.com/microsoft/FLAML/blob/816a82a1155b4de4705b21a615ccdff67c6da379/.github/workflows/python-package.yml#L54-L59) output is two 0s.
- [x ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ x] I've made sure all auto checks have passed.


EgorKraevTransferwise:
The Keyerror test fail is a strange one, last time it surfaced on Ubuntu while Windows was fine, and disappeared from Ubuntu on next push without my changing anything, while Mark tried to reproduce it locally and failed. Will dig into it next week.

flaml/config.py
```diff
@@ -4,7 +4,7 @@
 """
 
 N_SPLITS = 5
-RANDOM_SEED = 1
+RANDOM_SEED = 2
```
markharley, line 7: Not going to block the PR for this, but odd we have release code with a fixed random seed

markharley (approved):

markharley:
Put a minor comment on but otherwise looks great

flaml/automl/state.py
```diff
@@ -242,17 +244,22 @@ def est_retrain_time(self, retrain_sample_size):
 
 
 class AutoMLState:
-    def _prepare_sample_train_data(self, sample_size: int):
+    def prepare_sample_train_data(self, sample_size):
```
int-chaos, line 247: keep type annotation

flaml/automl/automl.py
```diff
@@ -1215,6 +1214,7 @@ def fit(
         free_mem_ratio=0,
         metric_constraints=None,
         custom_hp=None,
+        time_col=None,
```
int-chaos, line 1217: Does it still default to first column if no `time_col` is passed it? this is crucial to keep user's current code working with new version

int-chaos: also, it would be great if a description is added to docstring

int-chaos: Update: I see that it defaults to first col in flaml/automl/task/time_series_task.py

flaml/automl/state.py
EgorKraevTransferwise: Done

flaml/automl/automl.py
EgorKraevTransferwise: Docstring description added

int-chaos (approved):

flaml/automl/task/task.py
```diff
@@ -345,11 +342,11 @@ def estimator_class_from_str(cls, estimator_name: str) -> "flaml.automl.ml.BaseE
         Raises:
             ValueError: The provided estimator_name has not been registered for this task type.
         """
-        if estimator_name in cls.estimators:
-            return cls.estimators[estimator_name]
+        if estimator_name in self.estimators:
+            return self.estimators[estimator_name]
```
sonichi, line 346: It looks that `self.estimators` is indeed called often. The performance issue thus exists. Try to make the import happen only once by storing the dict in a private class member and don't compute it every time.

flaml/automl/task/time_series_task.py
```diff
@@ -0,0 +1,508 @@
...
+            LGBM_TS,
+            ExtraTrees_TS,
+            CatBoost_TS,
+            Prophet,
+            Orbit,
+            ARIMA,
+            SARIMAX,
+            TemporalFusionTransformerEstimator,
+            HoltWinters,
+        )
```
sonichi, line 48: performance issue as mentioned above.

flaml/automl/task/time_series_task.py
```diff
@@ -0,0 +1,508 @@
...
+            pred_time += pred_time_i
+            if time.time() - start_time >= budget:
+                break
+        val_loss, metric = cv_score_agg_func(val_loss_folds, log_metric_folds)
+        n = total_fold_num
+        pred_time /= n
+        return val_loss, metric, train_time, pred_time
+
+    def default_estimator_list(self, estimator_list: List[str], is_spark_dataframe: bool) -> List[str]:
+        assert not is_spark_dataframe, "Spark is not yet supported for time series"
```
sonichi, line 430: @thinkall is this true in previous version?

flaml/automl/task/time_series_task.py
```diff
@@ -0,0 +1,508 @@
...
+        return estimator_list
+
+    def default_metric(self, metric: str) -> str:
+        if "auto" != metric:
+            return metric
+
+        if self.is_ts_forecast():
+            return "mape"
+        else:
+            raise ValueError("If this is not a TS forecasting task, this code should never have been called")
```
sonichi, line 474: use assert

sonichi (commented):
reviewed automl/ and task/

flaml/default/suggest.py
```diff
@@ -178,9 +178,10 @@ def suggest_hyperparams(task, X, y, estimator_or_predictor, location=None):
     """
     config = suggest_config(task, X, y, estimator_or_predictor, location=location, k=1)[0]
     estimator = config["class"]
-    model_class = get_estimator_class(task, estimator)
+    task = task_factory(task)
+    model_class = task.estimator_class_from_str(estimator)
     hyperparams = config["hyperparameters"]
-    model = model_class(task=task, **hyperparams)
+    model = model_class(task=task.name, **hyperparams)
```
sonichi, line 184: Is there a reason for this change? It seems to make it less general.

ZmeiGorynych: As discussed, previously `task` was a string not an object, so this code is still getting the string that it expects.

flaml/automl/task/time_series_task.py
ZmeiGorynych: Currently in the case of spark dataframes, we only allow [estimators that are specifically for spark](https://github.com/microsoft/FLAML/blob/9977a7aae1d5d2bb7e6c09c23ec93802a95f6bd8/flaml/automl/task/generic_task.py#L958), and currently none of these do time series (in fact I believe the only one there right now is `lgbm_spark: SparkLGBMEstimator`

flaml/automl/automl.py
```diff
@@ -918,7 +917,7 @@ def _decide_eval_method(self, eval_method, time_budget):
             ], "eval_method must be 'auto' or 'cv' for custom data splitter."
             assert self._state.X_val is None, "custom splitter and custom validation data can't be used together."
             return "cv"
-        if self._state.X_val is not None:
+        if self._state.X_val is not None and not isinstance(self._state.X_val, TimeSeriesDataset):
```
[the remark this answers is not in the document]
ZmeiGorynych: Fixed: in the time series case, `self._state.X_val` and `self._state.X_train` point to the same TimeSeriesDataset instance, so the check for whether validation data is supplied must be different.

flaml/automl/task/generic_task.py
```diff
@@ -79,20 +66,38 @@ class psSeries:
...
+            LGBMEstimator,
+            LRL1Classifier,
+            LRL2Classifier,
+            CatBoostEstimator,
+            ExtraTreesEstimator,
+            KNeighborsEstimator,
+            TransformersEstimator,
+            TransformersEstimatorModelSelection,
+            SparkLGBMEstimator,
+        )
```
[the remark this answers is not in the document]
ZmeiGorynych: Fixed

flaml/automl/ml.py
```diff
@@ -20,32 +23,11 @@
...
-    ARIMA,
-    SARIMAX,
-    HoltWinters,
-    TransformersEstimator,
-    TemporalFusionTransformerEstimator,
-    TransformersEstimatorModelSelection,
-    SparkLGBMEstimator,
-)
+
+from flaml.automl.model import TransformersEstimator
```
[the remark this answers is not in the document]
ZmeiGorynych: Done

flaml/automl/task/task.py
ZmeiGorynych: Done

flaml/automl/task/time_series_task.py
ZmeiGorynych: Done

flaml/automl/task/generic_task.py
```diff
@@ -490,85 +447,51 @@ def prepare_data(
...
-                        y_val = y_train_all[y_train_all["time_idx"] > training_cutoff].drop(columns=ids)
-                    else:
-                        num_samples = X_train_all.shape[0]
-                        assert period < num_samples, f"period={period}>#examples={num_samples}"
-                        split_idx = num_samples - period
-                        X_train = X_train_all[:split_idx]
-                        y_train = y_train_all[:split_idx]
-                        X_val = X_train_all[split_idx:]
-                        y_val = y_train_all[split_idx:]
+                    raise Exception("For a TS forecast task, this code should never be called")
```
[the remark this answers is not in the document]
ZmeiGorynych: Done

flaml/config.py
EgorKraevTransferwise: Reset it back to its original value of 1, tests still run fine

flaml/automl/task/time_series_task.py
sonichi: Thanks. In that case, let's say "Spark dataframe is not ..."

flaml/automl/task/task.py
sonichi:
```suggestion
        estimator = self.estimators.get(estimator_name)
        if estimator is not None:
            return estimator
```

test/automl/test_forecast.py
```diff
@@ -1,8 +1,12 @@
+import datetime
+
 import numpy as np
+import pandas as pd
+
 from flaml import AutoML
 
 
-def test_forecast_automl(budget=5, estimators_when_no_prophet=["arima", "sarimax", "holt-winters"]):
+def test_forecast_automl(budget=30, estimators_when_no_prophet=["arima", "sarimax", "holt-winters"]):
```
sonichi, line 9: Is it necessary to run the test for this long? We want to keep test time as short as having enough coverage.

test/automl/test_forecast.py
```diff
@@ -572,10 +576,140 @@ def smape(y_pred, y_test):
...
+    from flaml import AutoML
+
+    tgts = ["y"]
+    # tgt = "SERIES_SANCTIONS"
+
+    preds = {}
+    for tgt in tgts:
+        features = []  # [c for c in train_df.columns if "SERIES" not in c and c != time_col]
+
+        automl = AutoML(time_budget=60, metric="mae", task="ts_forecast", eval_method="cv")
```
sonichi, line 607: Is it necessary to run the test for this long? We want to keep test time as short as having enough coverage.

sonichi (commented):
Reviewed a few more files. automl/time_series is not reviewed.

EgorKraevTransferwise:
Re https://github.com/microsoft/FLAML/pull/989#discussion_r1202692990;
why is your version better? It seems more verbose while doing the same thing.

test/automl/test_forecast.py
EgorKraevTransferwise: Duration reset to 5s

test/automl/test_forecast.py
EgorKraevTransferwise: Duration reset to 5s

EgorKraevTransferwise:
As far as I can tell, [this single test fail ](https://github.com/microsoft/FLAML/actions/runs/5067742145/jobs/9099162650?pr=989)has nothing to do with this PR, all tests relevant to the PR pass

EgorKraevTransferwise:
Merged in the latest changes from main and nothing works anymore, @sonichi any chance you could take a look?
```
import pandas as pd
ModuleNotFoundError: No module named 'pandas'
```

test/automl/test_forecast.py
```diff
@@ -572,10 +576,140 @@ def smape(y_pred, y_test):
...
+        # plt.plot(val_df[time_col], pred)
+        # plt.legend(["actual", "predicted"])
+        # plt.show()
+
+    print("yahoo!")
+
+
+#
+#
+# def test_finance_df():
```
qingyun-wu, line 639: Is the inclusion of the commented code intended?

test/test_model.py
```diff
@@ -101,34 +99,15 @@ def test_prep():
...
-            ],
-        }
-    )
-    y = np.array([0, 1, 0, 1, 0, 0])
-    lgbm.predict(X[:2])
-    lgbm.fit(X, y, period=2)
-    lgbm.predict(X[:2])
-    print(lgbm.feature_names_in_)
-    print(lgbm.feature_importances_)
+    # What's the point of callin ARIMA without parameters, or calling predict before fit?
```
qingyun-wu, line 102: Is the inclusion of the commented code intended?

flaml/automl/task/time_series_task.py
```diff
@@ -0,0 +1,508 @@
...
+    Task,
+    get_classification_objective,
+    TS_FORECAST,
+    TS_FORECASTPANEL,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class TimeSeriesTask(Task):
```
qingyun-wu, line 30: Any specific reason why `TimeSeriesTask` doesn't inherit from `GenericTask`?

flaml/automl/task/generic_task.py
```diff
@@ -664,19 +590,19 @@ def decide_split_type(
...
-                int,  # NOTE: _decide_split_type is before kwargs is updated to fit_kwargs_by_estimator
-            ), f"missing a required integer 'period' for '{TS_FORECAST}' task."
-            if fit_kwargs.get("group_ids"):
-                # TODO (MARK) This will likely not play well with the task class
-                self.name = TS_FORECASTPANEL
-                assert isinstance(
-                    fit_kwargs.get("group_ids"), list
-                ), f"missing a required List[str] 'group_ids' for '{TS_FORECASTPANEL}' task."
-            return "time"
+        # elif self.is_ts_forecast():
```
qingyun-wu, line 593: It seems this part is moved to the `decide_split_type` func of `TimeSeriesTask`. Are we assuming an instance of `GenericTask` is never a ts forecast task (thus, `self.is_ts_forecast` is always False)?  Even if so, I think it is better to add a warning or assertion msg here instead of directly removing it.

flaml/automl/state.py
```diff
@@ -56,6 +57,7 @@ def __init__(
         self,
         learner_class,
         data_size,
+        data,
```
qingyun-wu, line 60: A minor issue: When `data` is provided, `data_size` is presumably not necessary (could be induced from the given data).

flaml/automl/task/time_series_task.py
EgorKraevTransferwise: Because there's no opportunity for code reuse. Only stuff that's task-specific belongs into the ***Task classes, so what's the point of inheritance?

test/test_model.py
EgorKraevTransferwise: It was meant as a remark to reviewers as to why the below lines in the test (calling ARIMA without parameters, and calling predict before fit) were removed. In my opinion both should just result in an exception.

flaml/automl/task/generic_task.py
EgorKraevTransferwise: > Are we assuming an instance of GenericTask is never a ts forecast task?

Yes, we are. Will add an assertion.

flaml/automl/task/time_series_task.py
EgorKraevTransferwise: Well, maybe we'd eventually want subclasses of eg TimeSeriesTask specific to different forecasting tasks, but certainly GenericTask is on a different branch of the hierarchy

test/automl/test_forecast.py
EgorKraevTransferwise: No, will remove, thanks.

flaml/automl/state.py
EgorKraevTransferwise: Fixed.

qingyun-wu (approved):
I didn't take a very careful look at the files in the time_series folder. The other parts looks good to me now! Thank you!

test/test_model.py
sonichi: The code was there for code coverage. Removing will cause drop in code coverage.
Also, why is the LGBM_TS test removed?

ZmeiGorynych: For code coverage, would it not make more sense to have another test in test_forecast.py that calls each TS model in turn on a tiny task? Likewise, I removed the lgbm_ts test as I felt it was replicating the tests made in test_forecast_py. Can put it back in if you like, but would it not make more sense to instead add the for-loop over all the TS models that the TSTask knows, in test_forecast.py, perhaps in a follow-up PR? :)

sonichi: Making a follow-up PR is OK. Don't decrease code coverage in this PR.

sonichi: Making a follow-up PR is OK. Don't decrease code coverage in this PR.

EgorKraevTransferwise: Added a test in test_forecast.py that does a quick test fit for each of the available estimators. That fixes and improves coverage.

sonichi: Thanks. That's good. But the code below still covers code path which was not covered by the added code.

flaml/automl/time_series/ts_model.py
```diff
@@ -0,0 +1,760 @@
...
+            if len(self.regressors):
+                exog = self._preprocess(X[self.regressors])
+                forecast = self._model.predict(start=start, end=end, exog=exog.values, **kwargs)
+            else:
+                forecast = self._model.predict(start=start, end=end, **kwargs)
+        else:
+            raise ValueError(
+                "X needs to be either a pandas Dataframe with dates as the first column"
+                " or an int number of periods for predict()."
+            )
```
sonichi, line 344: This is an example of code path covered by the commented test code.

flaml/automl/task/time_series_task.py
int-chaos: @EgorKraevTransferwise @ZmeiGorynych seems like this was missed. please make this change.

ZmeiGorynych: What change? Spark is not supported yet for TS and we say so in the assert.

flaml/automl/time_series/ts_model.py
ZmeiGorynych: OK. put the test back in.

flaml/automl/task/task.py
EgorKraevTransferwise: Why? I find my formulation more readable, and the difference in speed will be in the single-digit milliseconds; if this was called thousands of times, we might want to count milliseconds, but this is called once per estimator evaluation, so the speed improvement will be negligible.

sonichi (commented):
Is the ts_forecast notebook tested with this PR?

test/test_model.py
```diff
@@ -110,26 +109,6 @@ def test_prep():
...
-                datetime(1900, 8, 9),
-            ],
-        }
-    )
-    y = np.array([0, 1, 0, 1, 0, 0])
-    lgbm.predict(X[:2])
-    lgbm.fit(X, y, period=2)
-    lgbm.predict(X[:2])
-    print(lgbm.feature_names_in_)
-    print(lgbm.feature_importances_)
```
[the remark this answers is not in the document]
EgorKraevTransferwise: Why? Why the special treatment for just these properties of just this estimator?

test/test_model.py
```diff
@@ -110,26 +109,6 @@ def test_prep():
...
-                datetime(1900, 3, 4),
-                datetime(1900, 3, 4),
-                datetime(1900, 7, 2),
-                datetime(1900, 8, 9),
-            ],
-        }
-    )
-    y = np.array([0, 1, 0, 1, 0, 0])
-    lgbm.predict(X[:2])
-    lgbm.fit(X, y, period=2)
```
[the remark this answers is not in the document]
sonichi: The repeating dates are for the deduplication feature I suppose. Not sure about the irregular space. @int-chaos Could you comment?

flaml/automl/task/time_series_task.py
```diff
@@ -0,0 +1,523 @@
...
+
+    duplicates = X.duplicated()
+
+    if any(duplicates):
+        logger.warning("Duplicate timestamp values found in timestamp column. " f"\n{X.loc[duplicates, X][time_col]}")
+        X = X.drop_duplicates()
+        logger.warning("Removed duplicate rows based on all columns")
+        assert (
+            X[[X.columns[0]]].duplicated() is None
+        ), "Duplicate timestamp values with different values for other columns."
```
sonichi, line 521: The coverage of this part of the code gets removed by the change in the test code.

test/test_model.py
```diff
@@ -101,31 +99,34 @@ def test_prep():
...
     X = DataFrame(
         {
             "A": [
-                datetime(1900, 2, 3),
-                datetime(1900, 3, 4),
-                datetime(1900, 3, 4),
+                datetime(1900, 3, 1),
+                datetime(1900, 3, 2),
+                datetime(1900, 3, 3),
                 datetime(1900, 3, 4),
```
sonichi, line 118:
```suggestion
                datetime(1900, 3, 4),
                datetime(1900, 3, 4),
                datetime(1900, 3, 4),
```

PR #1000, opened 2023-04-19
fix zerodivision

liususan091219:

Fix the bug https://github.com/microsoft/FLAML/actions/runs/4736844088/jobs/8408907423?pr=999

## Checks

- [ ] I've used [pre-commit](https://microsoft.github.io/FLAML/docs/Contribute#pre-commit) to lint the changes in this PR, or I've made sure [lint with flake8](https://github.com/microsoft/FLAML/blob/816a82a1155b4de4705b21a615ccdff67c6da379/.github/workflows/python-package.yml#L54-L59) output is two 0s.


flaml/automl/model.py
```diff
@@ -1169,19 +1169,19 @@ def predict(self, X, **pred_kwargs):
 
         kwargs = {} if self._task not in NLG_TASKS else {"metric_key_prefix": "predict"}
         try:
-            predictions = new_trainer.predict(test_dataset, **kwargs)
+            predictions = new_trainer.predict(test_dataset, **kwargs).predictions
         except ZeroDivisionError:
             logger.warning("Zero division error appeared in HuggingFace Transformers.")
             predictions = np.array([0] * len(test_dataset))
```
sonichi, line 1175: no return in this case?

thinkall (changes requested):
Could you add a test case to cover this?

thinkall:
https://github.com/microsoft/FLAML/actions/runs/4738979125/jobs/8413392652?pr=1000#step:14:429

@sonichi Do we need to care about this?
```
FAILED test/automl/test_classification.py::TestClassification::test_parallel_xgboost - ValueError: Ray component dashboard_agent_http is trying to use a port number 52365 that is used by other components.
```

sonichi:
> https://github.com/microsoft/FLAML/actions/runs/4738979125/jobs/8413392652?pr=1000#step:14:429
> 
> @sonichi Do we need to care about this?
> 
> ```
> FAILED test/automl/test_classification.py::TestClassification::test_parallel_xgboost - ValueError: Ray component dashboard_agent_http is trying to use a port number 52365 that is used by other components.
> ```

I guess that's transient. But this error needs to be addressed: https://github.com/microsoft/FLAML/actions/runs/4738979125/jobs/8413393167?pr=1000#step:13:177
which is what my comment suggested. @liususan091219

liususan091219:
> https://github.com/microsoft/FLAML/actions/runs/4738979125/jobs/8413392652?pr=1000#step:14:429
> 
> @sonichi Do we need to care about this?
> 
> ```
> FAILED test/automl/test_classification.py::TestClassification::test_parallel_xgboost - ValueError: Ray component dashboard_agent_http is trying to use a port number 52365 that is used by other components.
> ```



> Could you add a test case to cover this?

Yes, the test case was added in PR 926: https://github.com/microsoft/FLAML/pull/926/files, but later found that PR wasn't complete, this PR is meant to fix it.

flaml/automl/model.py
sonichi: where is the fix? I don't see any change.

sonichi (approved):

thinkall (approved):

PR #1008, opened 2023-04-24
update max_spark_parallelism to fit in auto-scale spark cluster

thinkall:

Update max_spark_parallelism to fit in auto-scale spark cluster. In a spark cluster with auto-scale enabled, the default parallelism is smaller than the actual value, we can overwrite it with FLAML_MAX_CONCURRENT.

## Checks


sonichi (commented):
Is any documentation change needed because the expected behavior will change?
Also, how do we know whether the cluster is auto-scale? The code doesn't seem to check that.

thinkall:
> Is any documentation change needed because the expected behavior will change? Also, how do we know whether the cluster is auto-scale? The code doesn't seem to check that.

I checked the related doc, no need to adjust. The user faced part keeps the same. And we don't need to know whether the cluster is auto-scale. The user who sets the environment variable should knows it. 

In most cases, no need to set this env var, it's used to overwrite the value read from the spark conf. In auto-scale cluster, the detected number of executors will be smaller than the max number of executors it has. In the old implementation, we'll not overwrite the maximum value and we can't trigger the auto-scale (tested in our system); so we added it in this PR.

sonichi:
> > Is any documentation change needed because the expected behavior will change? Also, how do we know whether the cluster is auto-scale? The code doesn't seem to check that.
> 
> I checked the related doc, no need to adjust. The user faced part keeps the same. And we don't need to know whether the cluster is auto-scale. The user who sets the environment variable should knows it.
> 
> In most cases, no need to set this env var, it's used to overwrite the value read from the spark conf. In auto-scale cluster, the detected number of executors will be smaller than the max number of executors it has. In the old implementation, we'll not overwrite the maximum value and we can't trigger the auto-scale (tested in our system); so we added it in this PR.

Does the change have any untended effect on non-auto-scale cluster?

sonichi (approved):

levscaut (approved):

PR #1039, opened 2023-05-14
response filter

### LINKED ISSUE #1049: Allow setting use_docker=False in the human proxy agent.
https://github.com/microsoft/FLAML/blob/7de4eb347d2fb7d69117aae4bb921fa620fdf234/flaml/autogen/agent/human_proxy_agent.py#L11

sonichi:

* Support using a filter to select responses. This is a generalization of using assertions to filter code completions as in `code_utils.implement`, and paves the way for using an external model to rate responses.
* Documentation and test.
* Allow user proxy agent to set `use_docker=False`.
* Utility functions for creating config list.

## Related issue number

Close #1049 

## Checks

- I've used [pre-commit](https://microsoft.github.io/FLAML/docs/Contribute#pre-commit) to lint the changes in this PR (note the same in integrated in our CI checks).


test/openai/test_completion.py
```diff
@@ -18,6 +19,50 @@
 from flaml.autogen.math_utils import eval_math_responses, solve_problem
 
 
+def yes_or_no_filter(context, response, **_):
+    return context["yes_or_no_choice"] is False or any(
```
qingyun-wu, line 23:
```suggestion
    return context.get("yes_or_no_choice", False) is False or any(
```

flaml/autogen/oai/completion.py
```diff
@@ -727,6 +730,15 @@ def create(
         )
         ```
 
+            filter_func (Callable, Optional): A function that takes in the context, the config and the response and returns a boolean to indicate whether the response is valid. E.g.,
+
+        ```python
+        def yes_or_no_filter(context, config, response):
+            return context["yes_or_no_choice"] is False or any(
```
qingyun-wu, line 737:
```suggestion
            return context.get("yes_or_no_choice", False) is False or any(
```

test/openai/test_completion.py
```diff
@@ -18,6 +19,50 @@
...
+
+
+def test_filter():
+    try:
+        import openai
+    except ImportError as exc:
+        print(exc)
+        return
+    response = oai.Completion.create(
+        context={"yes_or_no_choice": True},
```
qingyun-wu, line 45: Should also revise the docstr for `context`? 
E.g., Change
"It needs to contain keys that are used by the prompt template."
to 
"It needs to contain keys that are used by the prompt template and filter_func."

skzhang1 (commented):
It looks good to me.

skzhang1 (approved):

qingyun-wu (approved):

PR #1048, opened 2023-05-19
Blogpost for adaptation in HumanEval

sonichi:

Create a blogpost for the case study with HumanEval.

## Checks

- I've used [pre-commit](https://microsoft.github.io/FLAML/docs/Contribute#pre-commit) to lint the changes in this PR (note the same in integrated in our CI checks).


website/blog/2023-05-18-GPT-adaptive-humaneval/index.mdx
```diff
@@ -0,0 +1,164 @@
...
+    >>> vowels_count("abcde")
+    2
+    >>> vowels_count("ACEDY")
+    3
+    """
+```
+
+Can we predict whether GPT-3.5-Turbo can solve this task or do we need to use GPT-4? My first guess is that GPT-3.5-Turbo can get it right because the instruction is fairly straightforward. Yet, it turns out that GPT-3.5-Turbo does not consistently get it right, if we only give it one chance. It's not obvious (but an interesting research question!) how to predict the performance without actually trying.
+
+Good news? We notice that some simple example test cases are provided in the docstr. If we already have a response generated by GPT-3.5-Turbo or GPT-4, we can use those test cases to filter wrong implementations, and either use a more powerful model or generate more responses, until the result passes the example test cases. Moreover, this step can be automated by asking GPT-3.5-Turbo to generate assertion statements from the examples given in the docstr (a simpler task where we can place our bet) and executing the code.
```
qingyun-wu, line 43:
```suggestion
Good news? We notice that some simple example test cases are provided in the docstr. If we already have a response generated by GPT-3.5-Turbo or GPT-4, we can use those test cases to filter wrong implementations, and either use a more powerful model or generate more responses with a cheaper model, until the result passes the example test cases. Moreover, this step can be automated by asking GPT-3.5-Turbo to generate assertion statements from the examples given in the docstr (a simpler task where we can place our bet) and executing the code.
```

qingyun-wu, line 43: Why "Good news?"

website/blog/2023-05-18-GPT-adaptive-humaneval/index.mdx
```diff
@@ -0,0 +1,164 @@
...
+
+## Solution
+
+Combining these observations, we can design a solution with two intuitive ideas:
+
+* Make use of auto-generated feedback to filter responses.
```
qingyun-wu, line 49: Can you explain more explicitly what is "auto-generated feedback"?

website/blog/2023-05-18-GPT-adaptive-humaneval/index.mdx
qingyun-wu, line 43: Regarding "If we already have a response generated by GPT-3.5-Turbo or GPT-4":
Shall "having responses generated by GPT-3.5-Turbo or GPT-4" be considered an inherent property of the setting you are concerned about or a hypothetical condition that will be taken care of in the solution (for example you first use GPT-3.5-Turbo or GPT-4 to generate some response)? If it is the latter case, does it matter how the responses are generated, using GPT-3.5-Turbo or GPT-4 or even other cheaper models?

sonichi: not necessarily with a cheaper model.

sonichi: For transition. I can replace with "What else can we do?"

qingyun-wu: ok, got it.

sonichi: Tried to address this. Could you check again?

website/blog/2023-05-18-GPT-adaptive-humaneval/index.mdx
sonichi: Tried to explain.

qingyun-wu (approved):
Thanks! Looks good to me now.

liususan091219 (approved):

PR #1093, opened 2023-06-27
support string alg in tune

skzhang1:
Support specifying search algorithm by passing the corresponding string in tune. 

## Related issue number
N/A

## Checks

### DIFF
--- a/flaml/tune/searcher/flow2.py
+++ b/flaml/tune/searcher/flow2.py
@@ -317,6 +317,7 @@ class FLOW2(Searcher):
         metric: Optional[str] = None,
         mode: Optional[str] = None,
         config: Optional[Dict] = None,
+        lexico_objectives: Optional[Dict] = None,
     ) -> bool:
         if metric:
             self._metric = metric
@@ -331,6 +332,14 @@ class FLOW2(Searcher):
             self.space = config
             self._space = flatten_dict(self.space)
             self._init_search()
+        if lexico_objectives:
+            if "modes" not in lexico_objectives.keys():
+                self.lexico_objectives["modes"] = ["min"] * len(self.lexico_objectives["metrics"])
+            for t_metric, t_mode in zip(self.lexico_objectives["metrics"], self.lexico_objectives["modes"]):
+                if t_metric not in self.lexico_objectives["tolerances"].keys():
+                    self.lexico_objectives["tolerances"][t_metric] = 0
+                if t_metric not in self.lexico_objectives["targets"].keys():
+                    self.lexico_objectives["targets"][t_metric] = -float("inf") if t_mode == "min" else float("inf")
         return True
 
     def update_fbest(
--- a/flaml/tune/searcher/flow2.py
+++ b/flaml/tune/searcher/flow2.py
@@ -317,7 +317,6 @@ class FLOW2(Searcher):
         metric: Optional[str] = None,
         mode: Optional[str] = None,
         config: Optional[Dict] = None,
-        lexico_objectives: Optional[Dict] = None,
     ) -> bool:
         if metric:
             self._metric = metric
@@ -332,15 +331,6 @@ class FLOW2(Searcher):
             self.space = config
             self._space = flatten_dict(self.space)
             self._init_search()
-        if lexico_objectives:
-            self.lexico_objectives = lexico_objectives
-            if "modes" not in lexico_objectives.keys():
-                self.lexico_objectives["modes"] = ["min"] * len(self.lexico_objectives["metrics"])
-            for t_metric, t_mode in zip(self.lexico_objectives["metrics"], self.lexico_objectives["modes"]):
-                if t_metric not in self.lexico_objectives["tolerances"].keys():
-                    self.lexico_objectives["tolerances"][t_metric] = 0
-                if t_metric not in self.lexico_objectives["targets"].keys():
-                    self.lexico_objectives["targets"][t_metric] = -float("inf") if t_mode == "min" else float("inf")
         return True
 
     def update_fbest(
--- a/flaml/tune/searcher/flow2.py
+++ b/flaml/tune/searcher/flow2.py
@@ -333,6 +333,7 @@ class FLOW2(Searcher):
             self._space = flatten_dict(self.space)
             self._init_search()
         if lexico_objectives:
+            self.lexico_objectives = lexico_objectives
             if "modes" not in lexico_objectives.keys():
                 self.lexico_objectives["modes"] = ["min"] * len(self.lexico_objectives["metrics"])
             for t_metric, t_mode in zip(self.lexico_objectives["metrics"], self.lexico_objectives["modes"]):
--- a/flaml/tune/tune.py
+++ b/flaml/tune/tune.py
@@ -345,7 +345,7 @@ def run(
                 return
     ```
         search_alg: An instance/string of the search algorithm
-            to be used. The same instance/string can be used for iterative tuning.
+            to be used. The same instance can be used for iterative tuning.
             e.g.,
 
     ```python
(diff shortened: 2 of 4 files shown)


flaml/tune/tune.py
```diff
@@ -344,8 +344,8 @@ def easy_objective(config):
                 # do cleanup operation here
                 return
     ```
-        search_alg: An instance of BlendSearch as the search algorithm
-            to be used. The same instance can be used for iterative tuning.
+        search_alg: An instance/string of the search algorithm
+            to be used. The same instance/string can be used for iterative tuning.
```
sonichi, line 348:
```suggestion
            to be used. The same instance can be used for iterative tuning.
```
My understanding is that this sentence is about using the same instance for iterative tuning.

flaml/tune/tune.py
```diff
@@ -504,15 +503,19 @@ def easy_objective(config):
             try:
                 import optuna as _
 
-                SearchAlgorithm = BlendSearch
+                SearchAlgorithm = BlendSearch if not search_alg else eval(search_alg)
```
sonichi, line 506: The use of `eval` is not safe. Could you use an alternative way?

flaml/tune/tune.py
```diff
@@ -564,6 +567,12 @@ def easy_objective(config):
             searcher.set_search_properties(metric, mode, config, **setting)
         else:
             searcher.set_search_properties(metric, mode, config)
+        if lexico_objectives:
+            assert search_alg.__class__.__name__ in [
+                "BlendSearch",
+                "CFO",
+            ], "If lexico_objectives is not None, search_alg must be CFO or BlendSearch."
+            search_alg.lexico_objectives = lexico_objectives
```
sonichi, line 575: Shall we implement this in `set_search_properties`? Otherwise we need to document this way of setting `lexico_objectives` somewhere. Conceptually I think it should be part of `set_search_properties`.

sonichi (commented):
Please add a test.

flaml/tune/tune.py
```diff
@@ -481,12 +481,11 @@ def easy_objective(config):
         else:
             logger.setLevel(logging.CRITICAL)
 
-    from .searcher.blendsearch import BlendSearch, CFO
+    from .searcher.blendsearch import BlendSearch, CFO, RandomSearch
```
qingyun-wu, line 484: RandomSearch is not used.

skzhang1: we may use it in eval(search_alg)

flaml/tune/tune.py
skzhang1: resolved

flaml/tune/tune.py
skzhang1: I add assertion to make it feasible.

flaml/tune/tune.py
skzhang1: Yes. I agree. I support it in CFO for now. Will make it work in BlendSearch in future PR.

flaml/tune/tune.py
```diff
@@ -481,12 +481,18 @@ def easy_objective(config):
...
-    if search_alg is None:
+        logger.warning("If lexico_objectives is not None, search_alg is forced to be CFO or BlendSearch.")
+    if search_alg is None or isinstance(search_alg, str):
+        if isinstance(search_alg, str):
+            assert search_alg in [
+                "BlendSearch",
+                "CFO",
+                "CFOCat",
+                "RandomSearch",
+            ], "The string of the search_alg must be feasible."
```
sonichi, line 495: "feasible" is not clear.

flaml/tune/tune.py
sonichi: First, I don't like the use of `eval`. Unless you get an approval from an experience Python engineer, I'm against it. Even `locals` is better.
Second, the code doesn't look correct in the try...except block. The try part tries to import optuna before using BlendSearch. If the choice is not BlendSearch, we shouldn't enter the try...except block.

flaml/tune/searcher/flow2.py
```diff
@@ -317,6 +317,7 @@ def set_search_properties(
         metric: Optional[str] = None,
         mode: Optional[str] = None,
         config: Optional[Dict] = None,
+        lexico_objectives: Optional[Dict] = None,
```
sonichi, line 320: why is this argument added?

test/tune/test_tune.py
```diff
@@ -20,6 +21,39 @@
...
+
+def _easy_objective(config):
+    # Hyperparameters
+    width, height = config["width"], config["height"]
+
+    for step in range(config["steps"]):
+        # Iterative training function - can be any arbitrary training procedure
+        _evaluation_fn(step, width, height)
+        # Feed the score back back to Tune.
+        print("Trial stopped", step)
```
sonichi, line 54: remove?

test/tune/test_tune.py
```diff
@@ -352,6 +386,64 @@ def evaluate_config_empty(config):
...
+        config=so_search_space,
+    )
+    # lexico tune
+    tune.run(
+        _BraninCurrin, search_alg="CFO", num_samples=10, config=mo_search_space, lexico_objectives=lexico_objectives
+    )
+
+    ## Passing search_alg through instance
+    so_bs = BlendSearch(time_budget_s=5, metric="mean_loss", mode="min")
+    mo_bs = CFO(time_budget_s=5)
```
sonichi, line 427: mo_bs is CFO? Add a note here if you plan to change CFO into BlendSearch in future.

flaml/tune/searcher/flow2.py
skzhang1: removed

test/tune/test_tune.py
skzhang1: agree

flaml/tune/tune.py
```diff
@@ -500,20 +514,26 @@ def easy_objective(config):
...
-            try:
-                import optuna as _
+        if lexico_objectives:
+            SearchAlgorithm = CFO
+            logger.info(
+                f"Using search algorithm {SearchAlgorithm.__name__} for lexicographic optimization. Note that when providing other search algorithms, we use CFO instead temporarily."
+            )
+            metric = lexico_objectives["metrics"][0] or DEFAULT_METRIC
+        else:
+            if not search_alg or search_alg == "BlendSearch":
```
sonichi, line 524: when `search_alg=="BlendSearch"`, we shouldn't automatically switch to CFO. We should raise an error if the import fails.

flaml/tune/tune.py
```diff
@@ -555,6 +579,12 @@ def easy_objective(config):
         ):
             search_alg.use_incumbent_result_in_evaluation = use_incumbent_result_in_evaluation
         searcher = search_alg.searcher if isinstance(search_alg, ConcurrencyLimiter) else search_alg
+        if lexico_objectives:
+            assert search_alg.__class__.__name__ in [
+                "CFO",
+            ], "If lexico_objectives is not None, the search_alg must be CFO for now."
```
sonichi, line 585: Such assertions and logging happen in different places: here, 519, 487. Could you add TODO notes to modify them later in case they are forgotten?

test/tune/test_tune.py
```diff
@@ -352,6 +379,65 @@ def evaluate_config_empty(config):
...
+        search_alg="BlendSearch",
+        metric="mean_loss",
+        mode="min",
+        num_samples=10,
+        config=so_search_space,
+    )
+    # lexico tune
+    tune.run(
+        _BraninCurrin, search_alg="CFO", num_samples=10, config=mo_search_space, lexico_objectives=lexico_objectives
+    )
```
qingyun-wu, line 416: Add another test with lexico_objectives and`search_alg= "BlendSearch"`

qingyun-wu (approved):

flaml/tune/tune.py
skzhang1: yes. resolved.

flaml/tune/tune.py
skzhang1: Resolved

jtongxin (approved):

sonichi (approved):

PR #1100, opened 2023-06-29
temp solution for joblib 1.3.0 issue

### LINKED ISSUE #1102: Parallel tuning with Spark is no longer possible if joblib isn't pinned to < 1.3.0
When `use_spark` is set to True, tune.py tries imports `register_spark` from `joblibspark` and [calls it](https://github.com/microsoft/FLAML/blob/e3ca95bf8a5cc9c86638dab8ef0eb44bc1f3ee6f/flaml/tune/tune.py#L627).

However, since joblib 1.3.0 was released yesterday, a class used by `register_spark` no longer exists, which causes any attempted use of parallel tuning with Spark to fail:

image

I've already created [an issue](https://github.com/joblib/joblib-spark/issues/49) in the joblib-spark GitHub repo, but I thought I should let you guys know.

thinkall:

To fix spark parallel training errors.

Temp solution for joblib 1.3.0 issue, no need once https://github.com/joblib/joblib-spark/pull/48 is merged

## Related issue number

Closes #1102 

## Checks

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,6 @@ __version__ = version["__version__"]
 
 install_requires = [
     "NumPy>=1.17.0rc1",
-    "joblib<1.3.0",  # temp solution for joblib 1.3.0 issue, no need once https://github.com/joblib/joblib-spark/pull/48 is merged
 ]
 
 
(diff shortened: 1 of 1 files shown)


levscaut (approved):

sonichi (changes requested):
Add it to the option which needs joblib. Don't add it here.

setup.py
```diff
@@ -15,6 +15,7 @@
 
 install_requires = [
     "NumPy>=1.17.0rc1",
+    "joblib<1.3.0",  # temp solution for joblib 1.3.0 issue, no need once https://github.com/joblib/joblib-spark/pull/48 is merged
```
qingyun-wu, line 18: Add this requirement only under the "spark" option?

qingyun-wu (approved):

sonichi (approved):

PR #1133, opened 2023-07-18
Json config list, agent refactoring and new notebooks

### LINKED ISSUE #1137: Whether UserProxyAgent's `auto_reply` should call `self.send`?
It might be better that the `auto_reply` function (https://github.com/microsoft/FLAML/blob/main/flaml/autogen/agent/user_proxy_agent.py#L185) in UserProxyAgent return a string and let the `receive` method call the `send` function.

It would be easier to manage what's being sent if only `receive` can call `send` when we want to inherit UserProxyAgent  and modify `receive` or/and `auto_reply`

### LINKED ISSUE #1139: dated API usage in autogen_agent_planning notebook?
https://github.com/microsoft/FLAML/blob/main/notebook/autogen_agent_planning.ipynb

```
def ask_planner(message):
    planner.receive(message, planner_user)
    return planner._oai_conversations[planner_user.name][-1]["content"]
```
The `planner.receive` is not a suggested API to use?

### LINKED ISSUE #1141: The (Assistant) Agent could have a function to update its config
I think this is useful because sometimes we may want to add new functions for the function_call

sonichi:

1. provide a utility function to load a config list from json. One can put all the endpoints in a single place and filter when needed.
2. update notebooks.
3. sanitize the handling of config list.
4. handle openml error in the tests. https://github.com/openml/OpenML/issues/1196
5. move is_termination_msg arg from UserProxyAgent to Agent and allow AssistantAgent to terminate the conversation.
6. create an AIUserProxyAgent which inherits UserProxyAgent and can make ai_reply like AssistantAgent. 
7. add a notebook for multi-agent multi-user applications.
8. other refactoring of agents.
9. add a notebook example for using IPython to execute code.

## Related issue number

close #1139 
close #1137 
close #1141 

## Checks


pcdeadeasy (approved):
Reviewed changes to is_termination_message implementation.

gagb (approved):
The AIUserProxyAgent is working as intended. The code needs documentation to set execution configuration to be set false.

PR #1149, opened 2023-07-26
RolePlay chat for GenericAgent

LittleLittleCloud:

This PR implements `MultiAgent RolePlay` chat for `GenericAgent`, plus a notebook example on how to use it.

## Related issue number
will create issue later

## Checks


flaml/autogen/agent/responsive_agent.py
```diff
@@ -260,11 +261,19 @@ def generate_reply(self, messages: List[Dict], default_reply: Union[str, Dict] =
...
         if self._code_execution_config is False:
             return default_reply if self.oai_config is False else self._oai_reply(messages)
-        code_blocks = extract_code(message["content"])
+        
+        # search from the most recent message to the oldest message for code blocks
+        for msg in reversed(messages):
+            code_blocks = extract_code(msg["content"])
+            print("code_blocks", code_blocks)
+            if len(code_blocks) == 1 and code_blocks[0][0] != UNKNOWN:
+                break
```
sonichi, line 270: This is problematic because (1) the same code will get re-executed if no new code is suggested. (2) if human prevents code execution in one turn, the code may get executed in later turn.

LittleLittleCloud: I see. However, simply checking the most recent message might also be problematic. For example, when in the `code writer` -> `code-reviewer` -> `code-runner` pattern.

To prevent 1), maybe we can simply check the "new messages" for a specific agent? AKA the messages after that specific agent's most recent message. To resolve (2), we might need another llm call to retrieve the running code from chatting history?

sonichi: I pushed a change to use llm to find code blocks when regex fails in #1154. That could be an alternative to the approach you use now.

sonichi:
I reviewed all the files except the notebook. One high-level observation is that the existing communication method via send() and receive() is not used. Instead, a few new methods like `role_play` are directly called. I wonder whether we could/should incorporate this design in our framework. @qingyun-wu

Please review #1154 as it will affect this PR.

sonichi (commented):
I had these comments yesterday. It seems that they were not posted.

LittleLittleCloud:
Closing because the major functionality of this PR is already in main.

PR #1165, opened 2023-08-01
consecutive auto reply, history, template, group chat, class-specific reply

### LINKED ISSUE #1164: improve control of consecutive auto reply
```[tasklist]
### Tasks
- [ ] reset the counter of consecutive_auto_reply in initiate_chat
- [ ] allow separate max_consecutive_auto_reply for different senders and configurable in initiate_chat
```

sonichi:

- [x] reset the counter of consecutive_auto_reply in initiate_chat
- [x] allow separate max_consecutive_auto_reply for different senders
- [x] option to clear conversation history in initiate_chat
- [x] chess notebook using these features
- [x] disallow format str template by default
- [x] print instantiated message
- [x] skeleton of groupchat
- [x] support register class-specific reply function
- [x] group chat notebook

## Related issue number

close #1164 

## Checks


JieyuZ2 (approved):

flaml/autogen/agentchat/groupchat.py
```diff
@@ -0,0 +1,177 @@
+import sys
+from typing import Dict, List, Optional, Union
+import random
+from .agent import Agent
+from .responsive_agent import ResponsiveAgent
+
+
+class BroadcastAgent(ResponsiveAgent):
```
BeibinLi, line 8: I really like this design. I only have some lint comments here. Feel free to ignore them. 


1. Maybe we can rename the "BroadcastAgent" as a "Room" or "AgentGroup" or something simpler. In the real world, participants are often in the same "room" (or Chat Group), and the message a participant said should be sent to everyone inside the room.  So, it contains the "agents" list, which is more intuitive.
2. Maybe, we can add another class "RoundRobinChatManager(ChatManagerAgent)" which asks each participant to speak one-by-one in a round fashion.

BeibinLi (commented):
LGTM

flaml/autogen/agentchat/responsive_agent.py
```diff
@@ -101,12 +101,26 @@ def __init__(
...
+        self._max_consecutive_auto_reply = (
             max_consecutive_auto_reply if max_consecutive_auto_reply is not None else self.MAX_CONSECUTIVE_AUTO_REPLY
         )
         self._consecutive_auto_reply_counter = defaultdict(int)
+        self._max_consecutive_auto_reply_dict = defaultdict(self.max_consecutive_auto_reply)
         self._function_map = {} if function_map is None else function_map
         self._default_auto_reply = default_auto_reply
+        self._class_specific_reply = []
+
+    def register_class_specific_reply(self, class_type, reply_func: Callable):
```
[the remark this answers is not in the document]
sonichi: Could you create an issue based on the comment? I won't be able to address them in this PR but the decorator suggestion is very good interesting. Agent's name is volatile but we can discuss it later.

flaml/autogen/agentchat/groupchat.py
```diff
@@ -0,0 +1,204 @@
+import sys
+from typing import Dict, List, Optional, Union
+import random
+from .agent import Agent
+from .responsive_agent import ResponsiveAgent
+
+
+class Room(ResponsiveAgent):
```
LittleLittleCloud, line 8: it's unclear on the expected behavior over the following example if `Room` is an `ResponsiveAgent`
- `Room.Send` and `Room.Receive` from an external agent
- `Room.Send` and `Room.Receive` from other room
- nested `Room` agent inside `Room`

flaml/autogen/agentchat/groupchat.py
```diff
@@ -0,0 +1,204 @@
...
+        # set the name to sender's name if the role is not function
+        if message["role"] != "function":
+            message["name"] = sender.name
+        # broadcast the message to all agents except the sender
+        for agent in self.agents:
+            if agent != sender:
+                self.send(message, agent)
+
+
+class ChatManagerAgent(ResponsiveAgent):
```
LittleLittleCloud, line 50: Maybe we can move
- `select_next_speaker`
- `participant_information`

into `ChatManagerAgent`?

The `Room` will have
- A `ChatManagerAgent`
- A group of `ResponsiveAgent` as participants

In each round, `ChatManagerAgent` will be responsible for selecting the next speaker from the participants. `ChatManagerAgent` will also be responsible for receiving the message from next speaker.

Therefore, No extra change would be required for `ResponsiveAgent` to be a participant in a group chat.

ekzhu:
I think the new chess player agent notebook looks good.

ekzhu (approved):

flaml/autogen/agentchat/groupchat.py
sonichi: renamed

flaml/autogen/agentchat/responsive_agent.py
```diff
@@ -373,27 +510,16 @@ def generate_reply(
...
-                return default_reply
-            # code_blocks, _ = find_code(messages, sys_msg=self._oai_system_message, **self.llm_config)
-            # if len(code_blocks) == 1 and code_blocks[0][0] == UNKNOWN:
-            #     return code_blocks[0][1]
-            return self._oai_reply(messages)
-        # try to execute the code
-        exitcode, logs = self.execute_code_blocks(code_blocks)
-        exitcode2str = "execution succeeded" if exitcode == 0 else "execution failed"
-        return f"exitcode: {exitcode} ({exitcode2str})\nCode output: {logs}"
+        final, reply = self._check_termination_and_human_reply(sender=sender)
```
thinkall, line 513: Could we run the registered functions before the default `_check_termination_and_human_reply`? In my case, I need to check the received message, but when human_input_mode=="ALWAYS", the `_check_termination_and_human_reply` will return final as True, and my registered function will not be executed. I still need to override `generate_reply`. But if we run the registered functions first, I can achieve my goal without override the `generate_reply`.

thinkall (commented):
LGTM with one suggestion.

flaml/autogen/agentchat/responsive_agent.py
sonichi: If you skip human input by pressing Enter, `_check_termination_and_human_reply` should return final as False unless the msg is a termination msg.

flaml/autogen/agentchat/groupchat.py
BeibinLi: Very good point. I suggested the name "Room" but neglected the issues with "Room" inside the "Room". Let's discuss offline.

PR #1176, opened 2023-08-04
Use decorator for "register_auto_reply"

BeibinLi:
Use decorator to simply the auto reply function registration.

## Checks

- [ ] I've included any doc changes needed for https://microsoft.github.io/FLAML/. See https://microsoft.github.io/FLAML/docs/Contribute#documentation to build and test documentation locally.
- [ ] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [ ] I've made sure all auto checks have passed.

PR #1177, opened 2023-08-06
Make auto reply method pluggable

sonichi:

Right now registering a new auto reply method is not entirely modular. We can register a new auto reply method but that method may require some state initialization in the constructor. That makes the effort of adding an auto reply method not a single operation. Ideally we should be able to use a single operation to register an auto reply method.

This PR enables that. One example is that we can give any ResponsiveAgent the ability to manage a group chat now, without needing to inherit the GroupChatManager class.

The change in this PR makes it easy to plug different capabilities into an agent without inheriting. It also explicitly separates the context needed for each auto reply method. 

Big thanks to @ekzhu @LittleLittleCloud @BeibinLi for their inspiration.

## Related issue number

#1176 and #1158 will be affected by this PR.

## Checks


flaml/autogen/agentchat/responsive_agent.py
```diff
@@ -110,12 +111,19 @@ def __init__(
...
-        self.register_auto_reply(Agent, self._check_termination_and_human_reply)
+        self.register_auto_reply(Agent, ResponsiveAgent.generate_oai_reply)
+        self.register_auto_reply(Agent, ResponsiveAgent.generate_code_execution_reply)
+        self.register_auto_reply(Agent, ResponsiveAgent.generate_function_call_reply)
+        self.register_auto_reply(Agent, ResponsiveAgent.check_termination_and_human_reply)
 
-    def register_auto_reply(self, class_type, reply_func: Callable, position: int = 0):
+    def register_auto_reply(
+        self,
+        class_type,
```
ekzhu, line 121: (Optional) If agent has unique IDs, it helps to register auto-reply by agent ID, because you may want to have different auto-reply functions for different agents of the same class. e.g., agents of the same class with different system prompts and functions may actually be very different. You can also have group-id to associate group of agents with the same ID. Something to think about not required for this PR.

notebook/autogen_agentchat_chess.ipynb
ekzhu, line 1: Looks good.

ekzhu (approved):
What's your thoughts on ID and Group ID for agents? See my comments earlier. Otherwise it looks good to me.

flaml/autogen/agentchat/responsive_agent.py
sonichi: Good suggestion. I'll add three more types of matching: str (name), Agent (instance), and Callable (any user-defined matching function)

sonichi: Added

qingyun-wu (approved):
Looks good to me. The pluggable auto reply is useful. The group chat example is also interesting. Shall we add examples about auto reply on the documentation page (could be done in a different PR)?

PR #1186, opened 2023-08-18
fix generate_reply when sender is None.

kevin666aa:

`generate_reply` allows either messages or sender is None. But current code doesn't handle situation when sender is None. This PR fixes this bug.

## Checks

### DIFF
--- a/flaml/autogen/agentchat/responsive_agent.py
+++ b/flaml/autogen/agentchat/responsive_agent.py
@@ -748,17 +748,19 @@ class ResponsiveAgent(Agent):
             str or dict or None: reply. None if no reply is generated.
         """
         assert messages is not None or sender is not None, "Either messages or sender must be provided."
-        if sender is not None:
-            for reply_func_tuple in self._reply_func_list:
-                reply_func = reply_func_tuple["reply_func"]
-                if exclude and reply_func in exclude:
-                    continue
-                if asyncio.coroutines.iscoroutinefunction(reply_func):
-                    continue
-                if self._match_trigger(reply_func_tuple["trigger"], sender):
-                    final, reply = reply_func(self, messages=messages, sender=sender, config=reply_func_tuple["config"])
-                    if final:
-                        return reply
+        if messages is None:
+            messages = self._oai_messages[sender]
+
+        for reply_func_tuple in self._reply_func_list:
+            reply_func = reply_func_tuple["reply_func"]
+            if exclude and reply_func in exclude:
+                continue
+            if asyncio.coroutines.iscoroutinefunction(reply_func):
+                continue
+            if sender is None or self._match_trigger(reply_func_tuple["trigger"], sender):
+                final, reply = reply_func(self, messages=messages, sender=sender, config=reply_func_tuple["config"])
+                if final:
+                    return reply
         return self._default_auto_reply
 
     async def a_generate_reply(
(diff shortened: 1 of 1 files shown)


flaml/autogen/agentchat/responsive_agent.py
```diff
@@ -748,17 +748,19 @@ def generate_reply(
...
+        if messages is None:
+            messages = self._oai_messages[sender]
+
+        for reply_func_tuple in self._reply_func_list:
+            reply_func = reply_func_tuple["reply_func"]
+            if exclude and reply_func in exclude:
+                continue
+            if asyncio.coroutines.iscoroutinefunction(reply_func):
+                continue
+            if sender is None or self._match_trigger(reply_func_tuple["trigger"], sender):
```
sonichi, line 760:
```suggestion
            if self._match_trigger(reply_func_tuple["trigger"], sender):
```
I suggest removing `sender is None` because some reply function may exclude this condition in its trigger.

sonichi (commented):
It's better to provide a use case for sender=None to justify this change, and a test case.

flaml/autogen/agentchat/responsive_agent.py
kevin666aa: I don't understand "some reply function may exclude `sender is None` in its trigger", can you explain more? 

A use case is when `generate_reply` is called by itself, we just use it to generate a reply instead of engaging in a conversation. So, the sender is None and the message is not empty. `_match_trigger` is only used to match senders, and we should allow execution when sender is already None.

sonichi: I mean the trigger condition can be `sender is not None`. Then the intended behavior is not to trigger that function when `sender is None`.

kevin666aa:
How about this? `_match_trigger` accepts None, and we register the default generate replies with trigger `[Agent, None]`.

flaml/autogen/agentchat/responsive_agent.py
```diff
@@ -748,17 +748,19 @@ def generate_reply(
...
+        if messages is None:
+            messages = self._oai_messages[sender]
+
+        for reply_func_tuple in self._reply_func_list:
+            reply_func = reply_func_tuple["reply_func"]
+            if exclude and reply_func in exclude:
+                continue
+            if asyncio.coroutines.iscoroutinefunction(reply_func):
+                continue
+            if self._match_trigger(reply_func_tuple["trigger"], sender):
```
sonichi, line 760: Could you check whether the signature and implementation of this function needs a change due to this PR?

test/autogen/agentchat/test_responsive_agent.py
```diff
@@ -154,6 +154,27 @@ def test_responsive_agent():
...
+        given_num = 10
+        return num_to_be_added + given_num
+
+    dummy_agent_2 = ResponsiveAgent(name="user_proxy", human_input_mode="TERMINATE", function_map={"add_num": add_num})
+    messsages = [{"function_call": {"name": "add_num", "arguments": '{ "num_to_be_added": 5 }'}, "role": "assistant"}]
+
+    # when sender is None, messages is provided
+    assert (
+        dummy_agent_2.generate_reply(messages=messsages, sender=None)["content"] == "15"
+    ), "generate_reply not working"
```
sonichi, line 168:
```suggestion
    ), "generate_reply not working when sender is None"
```

test/autogen/agentchat/test_responsive_agent.py
```diff
@@ -154,6 +154,27 @@ def test_responsive_agent():
...
+
+    # when sender is provided, messages is None
+    dummy_agent_1 = ResponsiveAgent(name="dummy_agent_1", human_input_mode="ALWAYS")
+    dummy_agent_2._oai_messages[dummy_agent_1] = messsages
+    assert (
+        dummy_agent_2.generate_reply(messages=None, sender=dummy_agent_1)["content"] == "15"
+    ), "generate_reply not working"
```
sonichi, line 175:
```suggestion
    ), "generate_reply not working when messages is None"
```

sonichi (commented):
This way works.

flaml/autogen/agentchat/responsive_agent.py
kevin666aa: Updated comments.

sonichi (commented):
almost ready.

sonichi (approved):

skzhang1 (approved):

PR #1195, opened 2023-08-23
fix _append_oai_message

kevin666aa:

Fix bugs in function `_append_oai_message`

## Checks

### DIFF
--- a/flaml/autogen/agentchat/responsive_agent.py
+++ b/flaml/autogen/agentchat/responsive_agent.py
@@ -260,7 +260,7 @@ class ResponsiveAgent(Agent):
 
         Args:
             message (dict or str): message to be appended to the ChatCompletion conversation.
-            role (str): role of the message, can be "assistant" or "function".
+            role (str): role of the message, can be "assistant", "user" or "function", this will overwrite the `role` in the message.
             conversation_id (Agent): id of the conversation, should be the recipient or sender.
 
         Returns:
@@ -268,11 +268,11 @@ class ResponsiveAgent(Agent):
         """
         message = self._message_to_dict(message)
         # create oai message to be appended to the oai conversation that can be passed to oai directly.
-        oai_message = {k: message[k] for k in ("content", "function_call", "name", "context") if k in message}
-        if "content" not in oai_message and "function_call" not in oai_message:
+        oai_message = {k: message[k] for k in ("content", "function_call", "name", "role") if k in message}
+        # 'content' is a required field
+        if "content" not in oai_message or role not in ("assistant", "user", "function"):
             return False
-
-        oai_message["role"] = "function" if message.get("role") == "function" else role
+        oai_message["role"] = role
         self._oai_messages[conversation_id].append(oai_message)
         return True
 


flaml/autogen/agentchat/responsive_agent.py
```diff
@@ -260,19 +260,19 @@ def _append_oai_message(self, message: Union[Dict, str], role, conversation_id:
...
             conversation_id (Agent): id of the conversation, should be the recipient or sender.
 
         Returns:
             bool: whether the message is appended to the ChatCompletion conversation.
         """
         message = self._message_to_dict(message)
         # create oai message to be appended to the oai conversation that can be passed to oai directly.
-        oai_message = {k: message[k] for k in ("content", "function_call", "name", "context") if k in message}
-        if "content" not in oai_message and "function_call" not in oai_message:
+        oai_message = {k: message[k] for k in ("content", "function_call", "name", "role") if k in message}
```
sonichi, line 271: please don't remove "context".

flaml/autogen/agentchat/responsive_agent.py
```diff
@@ -260,19 +260,19 @@ def _append_oai_message(self, message: Union[Dict, str], role, conversation_id:
...
+        # 'content' is a required field
+        if "content" not in oai_message or role not in ("assistant", "user", "function"):
             return False
-
-        oai_message["role"] = "function" if message.get("role") == "function" else role
+        oai_message["role"] = role
```
sonichi, line 275: when the role is "user" and oai_message["role"]="function", do we want to overwrite oai_message["role"]?

sonichi (commented):
I have some questions.

flaml/autogen/agentchat/responsive_agent.py
kevin666aa: Yes. One purpose of this function is to change roles for some messages
However, the "role" is set as an required argument now, I think we can set it to optional and overwrite only when it is passed in.

sonichi: Then we need to change the part where this function is called. Otherwise, with your change, the function call feature will not work.

flaml/autogen/agentchat/responsive_agent.py
```diff
@@ -268,11 +270,16 @@ def _append_oai_message(self, message: Union[Dict, str], role, conversation_id:
...
         message = self._message_to_dict(message)
         # create oai message to be appended to the oai conversation that can be passed to oai directly.
-        oai_message = {k: message[k] for k in ("content", "function_call", "name", "role") if k in message}
+        oai_message = {k: message[k] for k in ("content", "function_call", "name", "role", "context") if k in message}
         # 'content' is a required field
-        if "content" not in oai_message or role not in ("assistant", "user", "function"):
+        if "content" not in oai_message:
+            return False
+        if role is not None:
+            oai_message["role"] = role
```
sonichi, line 278: This will still change "function" role to "user" and fail the function call feature.

flaml/autogen/agentchat/responsive_agent.py
```diff
@@ -268,11 +270,16 @@ def _append_oai_message(self, message: Union[Dict, str], role, conversation_id:
...
+        elif oai_message.get("role", "") not in ("assistant", "user", "function"):
```
sonichi, line 279: what about "system"?

flaml/autogen/agentchat/responsive_agent.py
```diff
@@ -262,19 +264,27 @@ def _append_oai_message(self, message: Union[Dict, str], role, conversation_id:
...
-        oai_message = {k: message[k] for k in ("content", "function_call", "name", "context") if k in message}
-        if "content" not in oai_message and "function_call" not in oai_message:
+        oai_message = {k: message[k] for k in ("content", "function_call", "name", "role", "context") if k in message}
+        # 'content' is a required field
+        if "content" not in oai_message:
+            return False
+        if role is not None:
+            if oai_message["role"] is "function" and role is not "function":
+                print(f"Warning: Attempt to overwrite role 'function' with {role}. Rejected.")
+                return False
```
sonichi, line 282: This will make the result to be regarded as invalid and prevent proper processing of function call results.

kevin666aa: Thank you for pointing out. I have revised it.

flaml/autogen/agentchat/responsive_agent.py
```diff
@@ -262,19 +264,27 @@ def _append_oai_message(self, message: Union[Dict, str], role, conversation_id:
...
         # create oai message to be appended to the oai conversation that can be passed to oai directly.
-        oai_message = {k: message[k] for k in ("content", "function_call", "name", "context") if k in message}
-        if "content" not in oai_message and "function_call" not in oai_message:
+        oai_message = {k: message[k] for k in ("content", "function_call", "name", "role", "context") if k in message}
+        # 'content' is a required field
+        if "content" not in oai_message:
+            return False
+        if role is not None:
+            if oai_message.get("role", "") == "function" and role != "function":
+                print(f"Warning: Attempt to overwrite role 'function' with {role}. Rejected.")
```
sonichi, line 281: Have you tested? This will insert a warning in the conversation will may confuse readers. I don't know why we want to print this message.

kevin666aa: You are right, I will remove the warning add add a comment.

flaml/autogen/agentchat/responsive_agent.py
```diff
@@ -262,19 +264,25 @@ def _append_oai_message(self, message: Union[Dict, str], role, conversation_id:
 
         Args:
             message (dict or str): message to be appended to the ChatCompletion conversation.
-            role (str): role of the message, can be "assistant" or "function".
+            role (str): role of the message, can be "assistant", "user" or "function", this will overwrite the `role` in the message only when original role is not "function".
```
sonichi, line 267:
```suggestion
            role (str): role of the message, can be "assistant", "user" or "function", this will overwrite the "role" in the message only when original role is not "function".
```

flaml/autogen/agentchat/responsive_agent.py
```diff
@@ -264,7 +264,7 @@ def _append_oai_message(
 
         Args:
             message (dict or str): message to be appended to the ChatCompletion conversation.
-            role (str): role of the message, can be "assistant", "user" or "function", this will overwrite the `role` in the message only when original role is not "function".
+            role (str): role of the message, can be "assistant", "user" or "function", this will overwrite the "role" in the message only when original role is not "function".
```
sonichi, line 267:
```suggestion
            role (str): role of the message. It can be "assistant", "user" or "function". It will be used to overwrite the "role" in the message only when original role is not "function".
```

sonichi (approved):

PR #1209, opened 2023-09-04
admin takeover in group chat

sonichi:

1. allow admin take over during group chat by keyboard interruption.
2. add a more complex group chat notebook example.

## Checks


flaml/autogen/agentchat/groupchat.py
```diff
@@ -12,6 +12,7 @@ class GroupChat:
     agents: List[Agent]
     messages: List[Dict]
     max_round: int = 10
+    admin_name: str = "Admin"  # the name of the admin agent
```
LittleLittleCloud, line 15: Maybe check if admin_name exists here? So you don't need to throw expection in run_chat

sonichi: how do we check it "here"?

LittleLittleCloud: My understand is `admin_name` must be the name of an agent in `agents`, Otherwise the `keyboardInterruption`  will be raised when human user tries to interrupt group chat? 

If my understand is correct it would be better to check if `admin_name` is legit(a.k.a is one of an `agents`) in constructor.

sonichi: `admin_name` is optional. There doesn't have to be an admin. Only when there is, we support using KeyboardInterruption to take over.

flaml/autogen/agentchat/groupchat.py
```diff
@@ -97,19 +98,29 @@ def run_chat(
...
             # broadcast the message to all agents except the speaker
-            for agent in config.agents:
+            for agent in groupchat.agents:
                 if agent != speaker:
                     self.send(message, agent, request_reply=False, silent=True)
-            if i != config.max_round - 1:
-                # speaker selection msg from an agent
-                speaker = config.select_speaker(speaker, self)
-                reply = speaker.generate_reply(sender=self)
+            if i != groupchat.max_round - 1:
```
pcdeadeasy, line 111: Please add a comment to explain what's being done here?

pcdeadeasy (commented):
Looks good to me. Although there can be a few more comments added to the notebook to explain the role of the different agents.

pcdeadeasy (approved):

flaml/autogen/agentchat/groupchat.py
sonichi: added.

qingyun-wu (approved):

PR #1222, opened 2023-09-19
Extract_code can detect single-line code.

BeibinLi:
A parameter is added. Default to False.
When True, it will also detect single-line code.

For instance, 
data = """

hello, how are you.

You can run `python main.py`

Or, you can write
```python
1 + 1 == 2
print(3)
```

```
Formated stuff here
and there
```

Otherwise, do `source download.sh` for it.

I would also want to extract variable `x`.

Is that ok?
"""

extract_code(data, detect_single_line_code=True)

We will get 
```
[('UNKNOWN', 'python main.py'),
 ('python', '1 + 1 == 2\nprint(3)'),
 ('UNKNOWN', 'Formated stuff here\nand there'),
 ('UNKNOWN', 'source download.sh'),
 ('UNKNOWN', 'x')]
```

## Checks

- [X ] I've made sure all auto checks have passed.

### DIFF
--- a/flaml/autogen/code_utils.py
+++ b/flaml/autogen/code_utils.py
@@ -26,24 +26,38 @@ def infer_lang(code):
     return "python"
 
 
-def extract_code(text: str, pattern: str = CODE_BLOCK_PATTERN) -> List[Tuple[str, str]]:
+def extract_code(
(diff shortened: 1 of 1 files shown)


BeibinLi:
Move to the autogen

PR #1259, opened 2023-12-02
Add ruff sort to pre-commit and sort imports in the library

### LINKED ISSUE #1252: Should we add isort to the pre-commit?
isort helps to organize the order of dependencies import, the current code is not formatted with isort. If a new PR is formatted with isort, there will be many changes in the order of dependencies import. The reordered code looks nice, but it also adds more work for the reviewers.

What about we use a PR to reformat the code, and also add isort to the pre-commit hooks?

glevv:

To maintain clean and lean imports

## Related issue number

Closes #1252

## Checks

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,6 +24,7 @@ select = [
 #    "D",  # see: https://pypi.org/project/pydocstyle
 #    "N",  # see: https://pypi.org/project/pep8-naming
 #    "S",  # see: https://pypi.org/project/flake8-bandit
+    "I",  # see: https://pypi.org/project/isort/
 ]
 ignore = [
     "E501",
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,7 +27,7 @@ repos:
     hooks:
     - id: black
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: v0.1.5
+    rev: v0.0.261
     hooks:
       - id: ruff
         args: ["--fix"]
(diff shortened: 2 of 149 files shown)


glevv:
I almost sure that some tests will start to fail, so there still some work to be done

glevv:
Should OpenAI tests also run?

thinkall (approved):
LGTM. Thank you very much @glevv.

thinkall:
> Should OpenAI tests also run?

This will be triggered later when approved.

glevv:
I'm not sure why OpenAI checks are failing.

thinkall:
> I'm not sure why OpenAI checks are failing.

@sonichi could you take a look at this? Thanks.

pyproject.toml
```diff
@@ -24,6 +24,7 @@ select = [
 #    "D",  # see: https://pypi.org/project/pydocstyle
 #    "N",  # see: https://pypi.org/project/pep8-naming
 #    "S",  # see: https://pypi.org/project/flake8-bandit
+    "I",  # see: https://pypi.org/project/isort/
```
Borda, line 27: this is the main change; the rest is just apply pre-commit :)
ref: https://github.com/microsoft/FLAML/issues/1252#issuecomment-1927245023

Borda (approved):

glevv:
I don't think it is possible to automerge without all the checks being green

Borda:
> I don't think it is possible to automerge without all the checks being green

That is right, but it seems that even all checks on mater are not green so this will need to wait until main branch is healthy =)

Borda:
@sonichi how about adding this [bot](https://github.com/marketplace/pre-commit-ci) which would help with all this manual linitng?

Issue #1262, opened 2023-12-12
Cannot predict with numpy inputs


leelew:
Hi,

I trained AutoML model for regression task. But when I use `model.predict()`, it returns TypeError. So how can I fix this problem?

Best,
Li

--------------------------------------------------------------------------------------------------
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/lilu/anaconda3/envs/am/lib/python3.10/site-packages/flaml/automl/automl.py", line 573, in predict
    X = self._state.task.preprocess(X, self._transformer)
  File "/home/lilu/anaconda3/envs/am/lib/python3.10/site-packages/flaml/automl/task/generic_task.py", line 637, in preprocess
    X = transformer.transform(X)
  File "/home/lilu/anaconda3/envs/am/lib/python3.10/site-packages/flaml/automl/data.py", line 394, in transform
    elif isinstance(X, DataFrame):
TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union

Programmer-RD-AI:
> try:
    import pandas as pd
    from pandas import DataFrame, Series
except ImportError:
    DataFrame = Series = pd = None

Could you check if you have installed pandas?
That may be the cause of the error...

PR #1307, opened 2024-05-22
Add logging import and logger initialization in spark module

Programmer-RD-AI:
This pull request adds an import statement for the logging module and initializes a logger object in the spark module of FLAML. The logger is used to issue a warning message if pandas is not installed. This ensures that users are notified if they attempt to use DataFrame and Series functionalities without having pandas installed, providing them with guidance on resolving the issue.

This change is necessary to improve user experience by providing clear and informative warning messages when pandas is not installed. It helps users understand why certain functionalities may not work as expected and guides them on how to resolve the issue.

## Related issue number

#1262


Borda (approved):
looks reasonable to me :)

PR #1360, opened 2024-09-30
fix: Cross validation process isn't always run to completion

### LINKED ISSUE #1349: Cross-validation process isn't always completed across all folds
Hi all,

I've found an issue where sometimes, when the time_budget is running out, the final model/s aren't evaluated across all n_splits requested by the user, but instead the process is ended early. 

For example, a user might want 10 fold cross-validation to be used, but actually, on the final model, only 4 fold cross validation is used. This means that on rare occasions, FLAML returns a best model (i.e. `automl.model`)  which was evaluated across less folds than the user requested. It's possible that were that model evaluated across all requested folds, it would no longer be the best model.

I can understand why FLAML ends early when the time budget is low, but it would be great if there was either an option to switch this off (i.e. once a model is selected, it is evaluated across all folds) or have the automl.model return the best _completed_ model, not just the model with the highest average validation loss.

This is a hard issue to replicate for two reasons:
1) Performance of FLAML differs on different machines
2) The default logging doesn't log the number of folds used.

I've done my best to make this issue reproducible by writing a custom logging function `cv_score_agg_func`, which logs an 'n_folds' attribute, showing the number of folds that model has been evaluated against.

This code leads to the below (attached) logs on my machine, which showcase the error (see the "n_folds" 4 on the final line). Others may need to tweak the time_budget to trigger a similar issue.

```
from flaml import AutoML
from sklearn import datasets

def cv_score_agg_func(val_loss_folds, log_metrics_folds):
    metric_to_minimize = sum(val_loss_folds)/len(val_loss_folds)
    metrics_to_log = None
    for single_fold in log_metrics_folds:
        if metrics_to_log is None:
            metrics_to_log = single_fold
        elif isinstance(metrics_to_log, dict):
            metrics_to_log = {k: metrics_to_log[k] + v for k, v in single_fold.items()}
        else:
            metrics_to_log += single_fold
    if metrics_to_log:
        n = len(val_loss_folds)
        metrics_to_log = (
            {k: v / n for k, v in metrics_to_log.items()}
            if isinstance(metrics_to_log, dict)
            else metrics_to_log / n
        )
        metrics_to_log["n_folds"] = n

    return metric_to_minimize, metrics_to_log

dic_data = datasets.load_iris(as_frame=True)  # numpy arrays
iris_data = dic_data["frame"]  # pandas dataframe data + target
automl = AutoML()
automl_settings = {
    "time_budget": 11,  # in seconds
    "metric": 'accuracy',
    "task": 'classification',
    "log_file_name": "incomplete_error.log",
    "log_type": "all",
    "eval_method": "cv",
    "n_splits":10,
    "cv_score_agg_func": cv_score_agg_func,
    "early_stop": False,
}
x_train = iris_data[["sepal length (cm)","sepal width (cm)", "petal length (cm)","petal width (cm)"]].to_numpy()
y_train = iris_data['target']
automl.fit(x_train, y_train, **automl_settings)
```

In the attached logs, it should be clear that the final model was not tested against all 10 folds.

[incomplete_error.log](https://github.com/user-attachments/files/16966351/incomplete_error.log)

In terms of package versions, I'm using FLAML 2.1.2, catboost 1.2.5, scikit-learn 1.5.0 and Python 3.12.0

dannycg1996:
See issue #1349.

Currently if the time budget has run out, the cross-validation process will exit early, and (in some cases) an incomplete model will be returned to the user as the best model.

This fix simply removes the possibility of the cross-validation process being exited early, ensuring that the training and testing of all models is run to completion. See full discussion of the tradeoffs [here](https://github.com/microsoft/FLAML/issues/1349)

So that users don't have incomplete models returned to them.

## Related issue number

Closes #1349 

## Checks

- [ ] I've used [pre-commit](https://microsoft.github.io/FLAML/docs/Contribute#pre-commit) to lint the changes in this PR (note the same in integrated in our CI checks).

### DIFF
--- a/flaml/automl/task/generic_task.py
+++ b/flaml/automl/task/generic_task.py
@@ -813,8 +813,6 @@ class GenericTask(Task):
             if is_spark_dataframe:
                 X_train.spark.unpersist()  # uncache data to free memory
                 X_val.spark.unpersist()  # uncache data to free memory
-            if budget and time.time() - start_time >= budget:
-                break
         val_loss, metric = cv_score_agg_func(val_loss_folds, log_metric_folds)
         n = total_fold_num
         pred_time /= n
--- a/flaml/automl/task/generic_task.py
+++ b/flaml/automl/task/generic_task.py
@@ -706,7 +706,6 @@ class GenericTask(Task):
             fit_kwargs = {}
         if cv_score_agg_func is None:
             cv_score_agg_func = default_cv_score_agg_func
-        start_time = time.time()
         val_loss_folds = []
         log_metric_folds = []
         metric = None


thinkall (commented):
Thanks for the PR, @dannycg1996 ! Could you please run pre-commit to fix the formatting issue? Line 120 is no longer needed.

https://github.com/microsoft/FLAML/actions/runs/11104831249/job/30849695446#step:6:1

dannycg1996:
Done!

thinkall (approved):

PR #1362, opened 2024-10-09
test: Adding tests to verify model reproducibility

### LINKED ISSUE #1361: [Issue]: Lack of Testing Around Reproducibility of Results of Classification and Regression Models
Hi all,

I've noticed that there are few FLAML issues centred on reproducibility. That is, if we take the best model which FLAML provides us with (`automl.model`), and retrain and test it on the same folds - we cannot always replicate the results. 
Could we create some unit tests around this? I had a quick glance through the codebase, and as far as I can tell, such tests don't currently exist.

I propose creating unit tests to verify the reproducibility of all in-built FLAML classification and regression models. An example of such a test for a CatBoostClassifier would be as follows:

1. Take an SKLearn dataset, such as the breast cancer dataset. Fix the resampling strategy to be cross-validation (CV), and restrict the estimators to only be catboost. Run it for say, 5 iterations. Run an AutoML instance with that configuration.
2. Take the best model which FLAML finds (which will be a CatBoostEstimator), and replicate the CV process i.e. train and test it on the exact same CV folds.
3. Assert that the reproduced score matches that provided by FLAML (using `pytest.approx`).
4. Repeat steps 2 and 3 with the underlying (i.e. the non FLAMLised model, `automl.model.model`), in this case a CatBoostClassifier.

To explain step 4, I'm making the assumption here that when fitted to the same folds, a FLAMLised model should produce identical results to the underlying SKLearn/CatBoost/XGBoost model. In my experience the two do not always return different results, as values are incorrectly mapped between them.

Based on of the currently open FLAML issues, quite a few of these tests would initially fail. I would initially comment such tests out. My intention with this issue is to be able to identify which models have issues, not to fix all reproducibility issues. Then in future, as reproducibility issues get fixed, the previously failing tests (which now pass) can be uncommented and made live. If they fail in future, we know that a bug has been introduced into the codebase. 

I should probably add that I'm happy to open a PR for this :)

I welcome any thoughts on the above - thanks all!

dannycg1996:

See associated issue #1361. There's a few issues around model reproducibility, and these tests will help identify where those issues are. I have commented out models which fail in the new tests, and these show where the reproducibility issues lie - largely around CatBoost, LGBM and Logistic Regression models.

I have added four separate tests in total, all of which aim to verify that we can reproduce the validation loss FLAML gives us, using the model which FLAML provides us with. The 4 tests deal with the following cases:
1. Reproduce the FLAML result, for the classification case, using the FLAMLised model.
2. Reproduce the FLAML result, for the classification case, using the underlying SKLearn/xgboost/catboost model.
3. Reproduce the FLAML result, for the regression case, using the FLAMLised model.
4. Reproduce the FLAML result, for the regression case, using the underlying SKLearn/xgboost/catboost model.

Tests 1 and 3, which deal with the FLAMLised model, are fairly self-explanatory (I hope). For the other tests (2 and 4) I had to define my own method on `conftest.py`, `evaluate_cv_folds_with_underlying_model`, which aims to mimic the FLAML cross-validation process. I couldn't directly call on `generic_task.evaluate_model_CV()` here, as that method doesn't work with non-FLAMLised estimators. The line `estimator.cleanup()` raises an error, for example.

Thanks for any feedback -please let me know of any changes which might be needed

## Related issue number

Closes #1361 

## Checks

### DIFF
--- a/test/automl/test_classification.py
+++ b/test/automl/test_classification.py
@@ -1,14 +1,28 @@
 import unittest
 from datetime import datetime
+from typing import Any, Dict, List, Union
 
 import numpy as np
 import pandas as pd
+import pytest
 import scipy.sparse
+from catboost import CatBoostClassifier, CatBoostRegressor, Pool
 from sklearn.datasets import load_breast_cancer
-from sklearn.model_selection import train_test_split
+from sklearn.metrics import f1_score
+from sklearn.model_selection import (
+    GroupKFold,
+    RepeatedStratifiedKFold,
+    StratifiedGroupKFold,
+    TimeSeriesSplit,
+    train_test_split,
+)
 
 from flaml import AutoML, tune
+from flaml.automl.ml import default_cv_score_agg_func, get_val_loss
 from flaml.automl.model import LGBMEstimator
+from flaml.automl.spark import psDataFrame
+from flaml.automl.spark.utils import len_labels
+from flaml.config import RANDOM_SEED
 
 
 class MyLargeLGBM(LGBMEstimator):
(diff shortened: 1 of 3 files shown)


dannycg1996:
@microsoft-github-policy-service agree company="Evotec"

thinkall (approved):
Thank you so much, @dannycg1996 ! LGTM.

PR #1364, opened 2024-10-15
fix: FLAML catboost metrics arent reproducible

### LINKED ISSUE #1317: n_estimators value on automl.model differs from value in logs (for CatBoost models)
Hi all,

The n_estimators value on the best model (`automl.model`) provided by FLAML does not seem to be set correctly for CatBoostClassifiers.

Example code here:

```
from flaml import AutoML
from sklearn import datasets

dic_data = datasets.load_iris(as_frame=True)  # numpy arrays
iris_data = dic_data["frame"]  # pandas dataframe data + target
automl = AutoML()
automl_settings = {
    "max_iter":2,
    "metric": 'accuracy',
    "task": 'classification',
    "log_file_name": "catboost_error.log",
    "log_type": "all",
    "estimator_list": ['catboost'],
    "eval_method": "cv",
}
x_train = iris_data[["sepal length (cm)","sepal width (cm)", "petal length (cm)","petal width (cm)"]].to_numpy()
y_train = iris_data['target']
automl.fit(x_train, y_train, **automl_settings)
print(automl.model.get_params())
```

The print statement logs the following for me:
{'early_stopping_rounds': 10, 'learning_rate': 0.09999999999999996, 'n_estimators': 33, 'thread_count': -1, 'verbose': False, 'random_seed': 10242048, 'task': <flaml.automl.task.generic_task.GenericTask object at 0x7f895f2b3830>, '_estimator_type': 'classifier'}

However, if I look into the actual [[catboost_error.log](https://github.com/user-attachments/files/16065817/catboost_error.log)], I can see that neither of the two estimators attempted had n_estimators = 33. They actually had n_estimators = 35 and n_estimators =57. Replicating the FLAML folds myself has shown that this n_estimators value should be 35, meaning that the logs are correct and automl.model is incorrect.

Furthermore, if I run `print(automl.model.model.get_all_params())` I get a dictionary which includes iterations=35. The [catboost documentation](https://catboost.ai/en/docs/references/training-parameters/common) shows that iterations is an alias of n_estimators, and whilst I haven't managed to pin down the exact cause of this issue, I believe it's tied in somewhere here. 

In terms of package versions, I'm using FLAML 2.1.2, catboost 1.2.5, scikit-learn 1.5.0 and Python 3.12.0

dannycg1996:
Note: Please don't merge without a discussion first

Currently it is impossible to reproduce the best_loss provided by FLAML, using the best model which FLAML provides - when that best model is a CatBoost model.

There are two important changes made in this PR:

1. Removal of `self.params[self.ITER_HP] = self._model.tree_count_` line on `model.py`
2. Removal of this if statement from `model.py`: ` if current_time + self._time_per_iter > self.deadline:
            return False`

The line highlighted in 1) is the cause of issue #1317. As I understand it, CatBoost estimators can be thought of as performing their own AutoML/optimisation process internally - they have early stopping, and when `use_best_model = True` they internally optimise the objective metric. The issue with the current code is that it overwrites the n_estimators value with the actual number of trees used, but doesn't switch off this AutoML functionality. Rather than providing the user with the correct model, we simply change the internal AutoML process. As such, we get different results when we retrain and test that model on the same folds.

2) is a little trickier, and needs some discussion. Essentially the callbacks change the behaviour of the models, when the time_budget remaining is less than the time which the model will take to train. This means that when we take the model FLAML provides, and refit it on the same folds but with no time budget involved, we get different behaviour and different results. I imagine that this callback was likely implemented as a means of ensuring that FLAML respects the time budget, but I personally think that results which aren't reproducible are not particularly useful. 

My main issue with 2) is that the `XGBoostEstimator` also utilises this same `check_resource_limits()`, so this change could have unintended consequences. On a similar note, I've realised that `XGBoostEstimator` and `CatBoostEstimators` call  `check_resource_limits` differently i.e.:
```
class XGBoostResourceLimit(BaseResourceLimit, TrainingCallback):
    def after_iteration(self, model, epoch, evals_log) -> bool:
        now = time.time()
        return not self.check_resource_limits(now, epoch, "xgb")


class CatBoostResourceLimit(BaseResourceLimit):
    def after_iteration(self, info) -> bool:
        now = time.time()
        return self.check_resource_limits(now, info.iteration, "cat")
```

Could you please explain why XGBoostResourceLimit returns `return not self.check_resource_limits(now, epoch, "xgb")` whilst CatBoostResourceLimit returns `self.check_resource_limits(now, info.iteration, "cat")` please?

It's the `not` part which I'm interested in.

Also worth mentioning is that we still assign `self._time_per_iter` in `check_resource_limits`. Can this code also be deleted, or is it still of use somewhere?

I'm fairly confident now that CatBoost results are now reproducible, but please let me know if you have any issues with my code.

## Related issue number

Closes #1317 

## Checks

### DIFF
--- a/flaml/automl/model.py
+++ b/flaml/automl/model.py
@@ -2090,6 +2090,8 @@ class CatBoostEstimator(BaseEstimator):
         if weight is not None:
             kwargs["sample_weight"] = weight
         self._model = model
+        # Commented-out line below incorrectly assigned n_estimators - see https://github.com/microsoft/FLAML/pull/1364
+        # self.params[self.ITER_HP] = self._model.tree_count_
         train_time = time.time() - start_time
         return train_time
 
--- a/flaml/automl/model.py
+++ b/flaml/automl/model.py
@@ -2753,7 +2753,7 @@ class BaseResourceLimit:
     def check_resource_limits(self, current_time, current_iteration, mllib):
         if (mllib == "xgb" and current_iteration == 0) or (mllib == "cat" and current_iteration == 1):
             self._time_per_iter = current_time - self.start_time
-        if current_time + self._time_per_iter > self.deadline:
+        if mllib != "cat" and current_time + self._time_per_iter > self.deadline:
             return False
         if psutil is not None and self.free_mem_ratio is not None:
             mem = psutil.virtual_memory()
--- a/flaml/automl/model.py
+++ b/flaml/automl/model.py
@@ -2090,7 +2090,6 @@ class CatBoostEstimator(BaseEstimator):
         if weight is not None:
             kwargs["sample_weight"] = weight
         self._model = model
-        self.params[self.ITER_HP] = self._model.tree_count_
         train_time = time.time() - start_time
         return train_time
 
@@ -2752,8 +2751,6 @@ class BaseResourceLimit:
     def check_resource_limits(self, current_time, current_iteration, mllib):
         if (mllib == "xgb" and current_iteration == 0) or (mllib == "cat" and current_iteration == 1):
             self._time_per_iter = current_time - self.start_time
-        if current_time + self._time_per_iter > self.deadline:
-            return False
         if psutil is not None and self.free_mem_ratio is not None:
             mem = psutil.virtual_memory()
             if mem.available / mem.total < self.free_mem_ratio:
--- a/flaml/automl/model.py
+++ b/flaml/automl/model.py
@@ -2753,6 +2753,8 @@ class BaseResourceLimit:
     def check_resource_limits(self, current_time, current_iteration, mllib):
         if (mllib == "xgb" and current_iteration == 0) or (mllib == "cat" and current_iteration == 1):
             self._time_per_iter = current_time - self.start_time
+        if current_time + self._time_per_iter > self.deadline:
+            return False
         if psutil is not None and self.free_mem_ratio is not None:
             mem = psutil.virtual_memory()
             if mem.available / mem.total < self.free_mem_ratio:
--- a/test/automl/test_classification.py
+++ b/test/automl/test_classification.py
@@ -427,7 +427,7 @@ class TestClassification(unittest.TestCase):
 @pytest.mark.parametrize(
     "estimator",
     [
-        # "catboost",
+        "catboost",
         "extra_tree",
         "histgb",
         "kneighbor",
@@ -488,7 +488,7 @@ def test_reproducibility_of_classification_models(estimator: str):
 @pytest.mark.parametrize(
     "estimator",
     [
-        # "catboost",
+        "catboost",
         "extra_tree",
         "histgb",
         "kneighbor",
--- a/test/automl/test_regression.py
+++ b/test/automl/test_regression.py
@@ -265,6 +265,53 @@ def test_reproducibility_of_regression_models(estimator: str):
         "metric": "r2",
         "keep_search_state": True,
         "skip_transform": True,
+        "retrain_full": True,
+    }
+    X, y = fetch_california_housing(return_X_y=True, as_frame=True)
+    automl.fit(X_train=X, y_train=y, **automl_settings)
+    best_model = automl.model
+    assert best_model is not None
+    config = best_model.get_params()
+    val_loss_flaml = automl.best_result["val_loss"]
+
+    # Take the best model, and see if we can reproduce the best result
+    reproduced_val_loss, metric_for_logging, train_time, pred_time = automl._state.task.evaluate_model_CV(
+        config=config,
+        estimator=best_model,
+        X_train_all=automl._state.X_train_all,
+        y_train_all=automl._state.y_train_all,
+        budget=None,
+        kf=automl._state.kf,
+        eval_metric="r2",
+        best_val_loss=None,
+        cv_score_agg_func=None,
+        log_training_metric=False,
+        fit_kwargs=None,
+        free_mem_ratio=0,
+    )
+    assert pytest.approx(val_loss_flaml) == reproduced_val_loss
+
+
+def test_reproducibility_of_catboost_regression_model():
+    """FLAML finds the best model for a given dataset, which it then provides to users.
+
+    However, there are reported issues around the catboost model - see here:
+    https://github.com/microsoft/FLAML/issues/1317
+    In this test we take the best catboost regression model which FLAML provided us, and then retrain and test it on the
+    same folds, to verify that the result is reproducible.
+    """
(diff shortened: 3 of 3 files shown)


flaml/automl/model.py
```diff
@@ -2752,8 +2751,6 @@ def __init__(self, start_time, deadline, free_mem_ratio):
     def check_resource_limits(self, current_time, current_iteration, mllib):
         if (mllib == "xgb" and current_iteration == 0) or (mllib == "cat" and current_iteration == 1):
             self._time_per_iter = current_time - self.start_time
-        if current_time + self._time_per_iter > self.deadline:
-            return False
```
thinkall, line 2756: This is needed.

flaml/automl/model.py
```diff
@@ -2090,7 +2090,6 @@ def fit(self, X_train, y_train, budget=None, free_mem_ratio=0, **kwargs):
         if weight is not None:
             kwargs["sample_weight"] = weight
         self._model = model
-        self.params[self.ITER_HP] = self._model.tree_count_
```
thinkall, line 2093: Let's comment this out instead of removing it.

thinkall (changes requested):
Thank you very much for the PR,  @dannycg1996 .

For your question on `XGBoostResourceLimit` and `CatBoostResourceLimit`, xgboost stops training if any callback function returns `True`, while catboost stops training if any callback function returns `False`.

https://catboost.ai/en/docs/concepts/parameter-tuning

https://optuna.readthedocs.io/en/v3.0.2/reference/generated/optuna.integration.CatBoostPruningCallback.html

https://github.com/catboost/catboost/blob/c114336b37f67fcd8af65fbcc9bb4684cde44f59/catboost/python-package/catboost/core.py#L7286-L7290

https://xgboost.readthedocs.io/en/stable/python/python_api.html#xgboost.callback.TrainingCallback


I think we should keep the `check_resource_limits`. There is a parameter `retrain_full` , if set to True, catboost result should be reproducible with commenting out Line 2093.

flaml/automl/model.py
dannycg1996: Done - added an explanatory comment too, linking to this PR.

flaml/automl/model.py
dannycg1996: Hi @thinkall,
I added the `retrain_full = True` to my own test case, and it didn't seem to affect my results. I looked at the documentation, which describes `retrain_full` as follows:
'_default=True | whether to retrain the selected model on the full training data when using holdout. True - retrain only after search finishes; False - no retraining; 'budget' - do best effort to retrain without violating the time budget._'
My tests use cross-validation, so it makes sense that `retrain_full` would have no effect. 

I re-added the if statement you highlighted above, and  attempted to see if I could find a FLAML configuration which returns a model which I am unable to replicate the result of. Unfortunately, I did find such a configuration. I've made this case into its own unit test on `test_regression.py` - it's called `test_reproducibility_of_catboost_regression_model`.
I'm not sure how reproducible this error will be when run on different hardware, so I'll paste one of the failing results I get below:
image

It is still my opinion that the above if statement should be removed - at least for the CatBoost case (I haven't tested XGBoost in depth yet). The runtime reduction from finishing Catboost training early is not, in my opinion, worth the trade-off of risking returning the wrong model to the user.

Please let me know your thoughts!

dannycg1996:
Hi @thinkall, thanks for the feedback!
I have added the deleted line back as a comment, with an explanatory line above it. However, I can still (in select cases) find errors with the reproducibility of CatBoostEstimators (even when the assigned `n_estimators` value is correct). I've made one such example into its own unit test - see my earlier comment. `

I should also add that removing the discussed if statement on `check_resource_limits` does fix this reproducibility issue, at least for the unit test I've written

thinkall (approved):
LGTM! Thank you @dannycg1996.

thinkall (commented):
image

dannycg1996:
Hi @thinkall, that does fix the issue! Thanks for the suggestion - I am happy to implement that change. I just want to raise a couple of things:

1. With your implementation, we lose the catboost callback entirely, including stopping training early when we don't have enough resources i.e. this code on `check_resource_limits`: 

```        if psutil is not None and self.free_mem_ratio is not None:
            mem = psutil.virtual_memory()
            if mem.available / mem.total < self.free_mem_ratio:
                return False
```
Personally I think this is fine, but thought I should double check that this isn't an issue.

2. This is more of a question. With your approach, we no longer have a need for the `CatBoostResourceLimit`. Can I delete this, to keep the code tidy?

Thanks again for any help :)

thinkall:
> Hi @thinkall, that does fix the issue! Thanks for the suggestion - I am happy to implement that change. I just want to raise a couple of things:
> 
> 1. With your implementation, we lose the catboost callback entirely, including stopping training early when we don't have enough resources i.e. this code on `check_resource_limits`:
> 
> ```
>             mem = psutil.virtual_memory()
>             if mem.available / mem.total < self.free_mem_ratio:
>                 return False
> ```
> 
> Personally I think this is fine, but thought I should double check that this isn't an issue.
> 
> 2. This is more of a question. With your approach, we no longer have a need for the `CatBoostResourceLimit`. Can I delete this, to keep the code tidy?
> 
> Thanks again for any help :)

Hi @dannycg1996 , thank you very much! I think we can just skip time budget for catboost:
image

dannycg1996:
> > Hi @thinkall, that does fix the issue! Thanks for the suggestion - I am happy to implement that change. I just want to raise a couple of things:
> > 
> > 1. With your implementation, we lose the catboost callback entirely, including stopping training early when we don't have enough resources i.e. this code on `check_resource_limits`:
> > 
> > ```
> >             mem = psutil.virtual_memory()
> >             if mem.available / mem.total < self.free_mem_ratio:
> >                 return False
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Personally I think this is fine, but thought I should double check that this isn't an issue.
> > 
> > 2. This is more of a question. With your approach, we no longer have a need for the `CatBoostResourceLimit`. Can I delete this, to keep the code tidy?
> > 
> > Thanks again for any help :)
> 
> Hi @dannycg1996 , thank you very much! I think we can just skip time budget for catboost: image

I've just made that change, thanks for the help!
I suspect that in some cases the callbacks in XGBoostEstimator will lead to similar issues with reproducibility. However, if I ever come across such issues, I'll tackle them in a separate PR :)
Thanks for the help - hopefully it's mergeable now!

thinkall (approved):

PR #1374, opened 2024-10-28
fix: Ensure Reproducibility of ElasticNetEstimator

### LINKED ISSUE #1373: [Bug]: Inconsistency in ElasticNet due to lack of random state
### Describe the bug

This is a nit-pick issue, but I extended the regression reproducibility tests locally to include ElasticNet.
I struggle to consistently reproduce the loss shown by FLAML, and I'm pretty sure this is due to not fixing a random state on the FLAML `ElasticNetEstimator`. Two runs of `test_reproducibility_of_underlying_regression_models[enet]` return four different results, for example:

Image

Image

Now of course these are very close, but there might be cases where the two results differ more. As part of this issue it would be good to fix a default random state to this ElasticNet estimator (as is done with many other estimators) and also to add 'enet' to `test_reproducibility_of_underlying_regression_models[enet]` and `test_reproducibility_of_regression_models[enet]`, so that we can be sure of the reproducibility of this estimator going forwards.

dannycg1996:
ElasticNetEstimator currently has no default random_state, so reproducing results is difficult. Described [here](https://github.com/microsoft/FLAML/issues/1373). Taking the best model and re-fitting it to the same folds will garner different results, which isn't ideal.

Ensures that model results are reproducible. 

## Related issue number

Closes #1373 

## Checks


thinkall (approved):
Nice! Thank you, @dannycg1996 .

PR #1377, opened 2024-11-11
Fix logging issues that can arise when importing flaml

kernelmethod:
Previously, FLAML would automatically set its own loglevel when imported, overriding any other logging settings that may previously have been set by users of the library. FLAML will now check whether a loglevel has been explicitly set before setting its own.

In addition, the `flaml.automl` warning is now emitted using `warnings` rather than `logging.warning`. This falls more in line with other FLAML submodules (e.g. `flaml.ml`).

Taken in conjunction, these two changes make it a lot easier for downstream users to suppress warnings emitted by FLAML upon importing the library, either by configuring logging prior to importing `flaml` or by using the somewhat more idiomatic
```
with warnings.catch_warnings(action="ignore"):
    import flaml
```
This change also helps to ensure that FLAML's defaults won't override users' explicit logging settings.

These changes make it easier for downstream users to configure FLAML's logging utilities as they see fit, and make it easier to suppress certain warnings that may occur when importing `flaml` for the first time.

## Related issue number

N/A

## Checks

### DIFF
--- a/flaml/__init__.py
+++ b/flaml/__init__.py
@@ -1,4 +1,5 @@
 import logging
+import warnings
 
 try:
     from flaml.automl import AutoML, logger_formatter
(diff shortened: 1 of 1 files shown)


thinkall (approved):
Thank you for the PR, @kernelmethod ! LGTM!

Issue #874, opened 2023-01-06
New tuning API in ray 2


sonichi:
There seems to be a big change in ray tune API: https://docs.ray.io/en/latest/tune/api_docs/execution.html#tuner
How does it affect flaml when using ray 2 as the backend?

@Yard1 your insight would be appreciated.

Yard1:
The only change that impacts FLAML is that `tune.run` will be replaced with `Tuner.fit()`. Aside from that, everything else will stay the same (and work in the same way). The transition shouldn't take more than 15 minutes.

sonichi:
Thanks. TODO: 
[ ] call `Tuner.fit()` when ray version >=2.0 in automl.py and tune.py.
[ ] add `Tuner.fit()` API in tune.py.

Please comment if I missed anything.

Yard1:
That sounds right!

bbudescu:
I tried running flaml tune with the latest version of ray (2.2.0), and it seems to work, as they probably kept the old API for backwards compatibility with code built against it. Is there currently any good reason for which one should refrain from upgrading to the latest version, and stick with the version suggested in the package's setup.py (1.13)?

Also, is this of any consequence when using just the tune section of flaml, rather than automl?

Yard1:
We are planning on deprecating the `tune.run` API eventually, but so far, it should work fine, that's right. For users of FLAML, I don't see a reason not to update.

PR #1392, opened 2025-01-09
Changed tune.report import for ray>=2

drwillcharles:

When using ray>=2, there is an import error for `tune.report`.
>AttributeError: module 'ray.tune' has no attribute 'report'

I used the Ray [migration guide](https://docs.google.com/document/d/1o-J5UAeH4M5gGN9BwdrCt1q-nLe58mp94xsmTaW0Flc/edit?tab=t.0) to update this and use the new report function if the ray version is not 1.x .

I see that in the FLAML codebase there are already a few workarounds like this so I think I can do something similar without needing a full migration to Ray 2.x
## Related issue number

Related to https://github.com/microsoft/FLAML/issues/874 - although I still think a full migration to Ray>=2 is needed in the future.
## Checks

- [ x] I've used [pre-commit](https://microsoft.github.io/FLAML/docs/Contribute#pre-commit) to lint the changes in this PR (note the same in integrated in our CI checks).
- [x ] I've included any doc changes needed for https://microsoft.github.io/FLAML/. See https://microsoft.github.io/FLAML/docs/Contribute#documentation to build and test documentation locally.
- [ x] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [x ] I've made sure all auto checks have passed.


drwillcharles:
@microsoft-github-policy-service agree company="Evotec"

thinkall (commented):
Thank you, @drwillcharles , for the PR!

thinkall:
Hi @drwillcharles , could you check the test errors? Looks like the change will lead to test timeout issue. Thanks.

drwillcharles:
It appears the test `_test_hf_data()` on `test_gpu.py` gets stuck in a loop because of an error in Ray. Strangely - this test is not even run because of the naming convention. (`_test_hf_data()` instead of `test_hf_data()`).

I have loosened the pydantic dependency and hope this solves the issue.

drwillcharles:
[OpenML](www.openml.org) is currently down so a lot of tests are getting connection errors. Will wait until this is back online before I resume development. 

From the tests that have run successfully, it appears there are still some environment issues to fix

thinkall (approved):

drwillcharles:
I've figured out that the jobs are timing out because Ray cannot find the appropriate resource 
image

There is a function `_test_hf_data` in test/test_gpu.py which I don't think is even picked up in the tests (because of the underscore). 
However, when running this file, it gets stuck in an infinite loop. I think before this just failed silently.

Can I delete this?

thinkall:
> I've figured out that the jobs are timing out because Ray cannot find the appropriate resource image
> 
> There is a function `_test_hf_data` in test/test_gpu.py which I don't think is even picked up in the tests (because of the underscore). However, when running this file, it gets stuck in an infinite loop. I think before this just failed silently.
> 
> Can I delete this?

Sorry for the late reply, @drwillcharles. It's interesting that this underscored test is picked up. For bypassing the test, I think we can  detect whether there is a GPU resource and skip the test if None. WDYT? Thanks.

drwillcharles:
No problem @thinkall. I did add a check for the GPU resource and even deleted the test to check. Turns out I was wrong and the problem appears to be `test_lrl2` on test_model.py.

Looking at the line `lr.fit(X, y, budget=1e-5)` it appears the test should just stop when the budget is exceeded (almost instantaneous). However, that appears not to be happening. I have not yet found a solution to this and it appears https://github.com/microsoft/FLAML/pull/1402 is having the same problem in the tests.

thinkall:
> No problem @thinkall. I did add a check for the GPU resource and even deleted the test to check. Turns out I was wrong and the problem appears to be `test_lrl2` on test_model.py.
> 
> Looking at the line [`lr.fit(X, y, budget=1e-5)`](https://github.com/microsoft/FLAML/blob/6d539298036b144b968dd70b1089fb9436eef4b1/test/test_model.py#L26) it appears the test should just stop when the budget is exceeded (almost instantaneous). However, that appears not to be happening. I have not yet found a solution to this and it appears #1402 is having the same problem in the tests.

Thank you for pointing out the root cause test, @drwillcharles . I fixed the tests in #1406 .

drwillcharles:
> > No problem @thinkall. I did add a check for the GPU resource and even deleted the test to check. Turns out I was wrong and the problem appears to be `test_lrl2` on test_model.py.
> > Looking at the line [`lr.fit(X, y, budget=1e-5)`](https://github.com/microsoft/FLAML/blob/6d539298036b144b968dd70b1089fb9436eef4b1/test/test_model.py#L26) it appears the test should just stop when the budget is exceeded (almost instantaneous). However, that appears not to be happening. I have not yet found a solution to this and it appears #1402 is having the same problem in the tests.
> 
> Thank you for pointing out the root cause test, @drwillcharles . I fixed the tests in #1406 .

Thanks @thinkall ! I have implemented your fix so hopefully these tests pass now 🤞

Issue #807, opened 2022-11-08
Isolate the ensemble part and expose it to users


sonichi:
Right now the ensemble part is not isolated from hyperparameter search. Sometimes the ensemble part fails due to OOM. Users would like to use a more powerful machine to finish the ensemble part using a pickled automl object. #801 

Suggestion: put the ensemble part in a separate function and expose it to users.

coffepowered:
Can you provide some hints on where you would start? I am evaluating if I'm able to contribute on this

sonichi:
> Can you provide some hints on where you would start? I am evaluating if I'm able to contribute on this

The ensemble part is located in https://github.com/microsoft/FLAML/blob/63d350d4c8cae5f5bdd686a97e737013fbd9641d/flaml/automl/automl.py#L3625-L3746

PR #1420, opened 2025-04-08
estimator function

007-WJR:
Open #807 Isolate the ensemble part and expose it to users


007-WJR:

Issue #1422, opened 2025-04-14
[Bug]: best_model_for_estimator returns inconsistent feature_importances_ compared to automl.model when model_history=True


Stickic-cyber:
### Describe the bug

While testing Issue #1398, I found an inconsistency related to feature_importances_:

When I manually set `model_history=True`, the call works and returns` feature_importances_`.

However, the `feature_importances_` from `automl.best_model_for_estimator(automl.best_estimator)` and from `automl.model` are not the same.

This leads me to suspect that `best_model_for_estimator` may not be returning the actual best model selected by AutoML.

### Steps to reproduce

```
from flaml import AutoML
import pandas as pd
import numpy as np

np.random.seed(41) 
n_samples = 50

X = pd.DataFrame({
    'feature1': np.random.rand(n_samples) * 10, 
    'feature2': np.random.randint(0, 5, n_samples),
    'feature3': np.random.normal(5, 2, n_samples)
})

y_train = 2 * X['feature1'] + 3 * X['feature2'] - 0.5 * X['feature3'] + np.random.randn(n_samples) * 2

settings = {
    "time_budget": 60,
    "estimator_list": ["extra_tree", "xgboost"],
    "task": "regression",
    "log_file_name": "test_ensemble.log",
    "seed": 41,
    "ensemble": False,
    "n_concurrent_trials": 1,
    "verbose": 1,
    "metric": "rmse",
    "mlflow_logging": True,
    "model_history": True
}

automl = AutoML()
automl.fit(X_train=X, y_train=y_train, **settings)

best_model = automl.best_model_for_estimator(automl.best_estimator)
print(automl.best_estimator)

feature_importances1 = best_model.feature_importances_ # 输出为None(当"model_history"为默认False时)
print("Feature importances1:", feature_importances1)
feature_importances2 = automl.model.feature_importances_ # 输出不为None
print("Feature importances2:", feature_importances2)
```

### Screenshots and logs

![Image](https://github.com/user-attachments/assets/5b8bba4b-fe09-43a2-b82c-90874fa48683)

### Additional Information

FLAML Version: 2.3.4
python Version: 3.11
Operating System: Windows

PR #1435, opened 2025-05-22
Add the detailed info for parameter 'verbose'

murunlin:

Add the detailed explanation for parameter 'verbose', which can control the logger level such as ERROR, WARNING messages. 

## Related issue number

Related to #1422 

## Checks

- [ ] I've used [pre-commit](https://microsoft.github.io/FLAML/docs/Contribute#pre-commit) to lint the changes in this PR (note the same in integrated in our CI checks).

### DIFF
--- a/flaml/automl/automl.py
+++ b/flaml/automl/automl.py
@@ -190,6 +190,9 @@ class AutoML(BaseEstimator):
             train_time_limit: A float of the training time constraint in seconds.
             verbose: int, default=3 | Controls the verbosity, higher means more
                 messages.
+                verbose=1: logger level = ERROR, CRITICAL + ERROR log messages will be shown.
+                verbose=2: logger level = WARNING, there will be CRITICAL + ERROR + WARNING log messages.
+                verbose=3: logger level = INFO, add INFO messages to the above.
             retrain_full: bool or str, default=True | whether to retrain the
                 selected model on the full training data when using holdout.
                 True - retrain only after search finishes; False - no retraining;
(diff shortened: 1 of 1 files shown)


thinkall (commented):
Why only `(shows ERROR + CRITICAL)` and `(all messages)`? I suggest remove both of them.

As I've tested, verbose 5 will not show all messages, but verbose > 5 will do. Please add info of verbose > 5.

Thanks.

thinkall (approved):

setup.py
```diff
@@ -40,7 +40,7 @@
             "xgboost>=0.90,<3.0.0",
             "scipy>=1.4.1",
             "pandas>=1.1.4",
-            "scikit-learn>=1.0.0",
+            "scikit-learn==1.3.1",
```
thinkall, line 43: This is not necessary and we should not stick to any specific version.

setup.py
```diff
@@ -57,7 +57,7 @@
             "scipy>=1.4.1",
             "pandas>=1.1.4,<2.0.0; python_version<'3.10'",
             "pandas>=1.1.4; python_version>='3.10'",
-            "scikit-learn>=1.0.0",
+            "scikit-learn==1.3.1",
```
thinkall, line 60: This is not necessary and we should not stick to any specific version.

.github/workflows/python-package.yml
```diff
@@ -85,7 +85,7 @@ jobs:
       - name: Test with pytest
         if: matrix.python-version != '3.10'
         run: |
-          pytest test/
+          pytest test/ --ignore=test/automl/test_mlflow.py
```
thinkall, line 88: We should not skip the test.

thinkall (changes requested):

PR #1452, opened 2025-07-31
Unpin Pydantic

OlivierBinette-UpStart:

`FLAML` currently uses a pinned version of Pydantic, version `1.10.9`. This makes it impossible to install `FLAML` alongside libraries that use other versions of Pydantic or Pydantic v2.

Unclear why that limitation is required. Pydantic is only imported in a single place within the whole of the package.

Changes:

- I the single import of `pydantic` to be compatible with both version 1 and version 2.
- I unpinned the Pydantic version in setup.py.

## Checks

- [] I've included any doc changes needed for https://microsoft.github.io/FLAML/. See https://microsoft.github.io/FLAML/docs/Contribute#documentation to build and test documentation locally.
- [] I've added tests (if relevant) corresponding to the changes introduced in this PR.
- [] I've made sure all auto checks have passed.


thinkall (approved):
Thank you very much for the PR. LGTM.

thinkall:
> @thinkall It seems like older versions of ray[tune] are not compatible with pydantic v2. I bumped it to `ray[tune]>=2.9` in setup.py, but that might be a more significant dependency change.

Hi @OlivierBinette-UpStart , ray 2 is not supported yet. Let's see will it work.

thinkall:
> @thinkall Oh you're totally right. We're using ray[tune], which is where the Pydantic dependency must be coming from.
> 
> Do you think FLAML can be compatible with newer versions of Ray?

The latest PR for supporting ray is #1392, but I'm not sure will it work for other modules or not. I think some changes will be needed.

PR #1508, opened 2026-01-21
Auto-Create Timestamps in `prettify_prediction()` When `test_data` is None

aditisingh02:

Currently, the `TimeSeriesDataset.prettify_prediction()` method in `flaml/automl/time_series/ts_data.py` throws a `NotImplementedError` when `test_data` is `None`:

```python
# TODO auto-create the timestamps for the time column instead of throwing
raise NotImplementedError("Need a non-None test_data for this to work, for now")
```

This is frustrating for users who want to make predictions without providing explicit test data timestamps, which is a common use case in time series forecasting.

**This PR implements automatic timestamp generation** by:

1. Using the training data's end date (`train_data[time_col].max()`) as the starting point
2. Generating future timestamps based on the dataset's inferred `frequency`
3. Supporting all input types: `np.ndarray`, `pd.Series`, and `pd.DataFrame`

### Example behavior after this change:

```python
# Before: NotImplementedError
# After: Automatically generates timestamps starting from training end + 1 period
y_pred = model.predict(steps=10)  # Works without explicit test_data!
```
## Related issue number
Closes #1507

## Checks


thinkall (commented):
Thank you so much for the PR, @aditisingh02 ! Could you please also add a test case for your changes?

thinkall (commented):
@aditisingh02 , thank you so much for the revision. Copilot's feedbacks are not always correct. For instance, pd.Timedelta actually can accept frequency as its second parameter:
https://pandas.pydata.org/docs/reference/api/pandas.Timedelta.html . Sorry for the mistakes made by the Copilot review agent.

Do you mind make some changes to the tests you've added? I'd prefer simplify the tests. Besides, I believe an E2E tests with model training and prediction can better showcase the improvements of this PR.

The last minor change I'd suggest is that putting the new tests in test_forecast.py instead of test_model.py.

thinkall (changes requested):
@aditisingh02 , thank you so much for the active revision. However, the agent you're using doesn't seem to have the ability to resolve the issue in your code. It's time to go with "human-in-the-loop" :-)

thinkall:
Hi @aditisingh02 , thank you very much for the PR. There are currently some test failures, and I don’t see a fix included yet. Additionally, PR #1536 already addresses the same issue, so I’m going to close this one.

You’re very welcome to submit new PRs in the future.

PR #1554, opened 2026-05-23
fix: prepare_data crashes with non shuffle split + sample_weight (#887, #1553)

### LINKED ISSUE #887: 'AutoMLState' has no attribute 'sample_weight_all' when split_type='time'
I'm struggling to find a minimal example for this to share, but I think I can describe the issue. I think I could find a minimal example if I better knew how to get into [this elif clause in the `_search` method](https://github.com/microsoft/FLAML/blob/cdc4201188565e74507c026f86abe540ade4cb72/flaml/automl/automl.py#L3746) at the end of fitting.

The only place I can see the `sample_weight_all` attribute being set is [in AutoML._prepare_data](https://github.com/microsoft/FLAML/blob/main/flaml/automl/automl.py#L1315) and that only happens when `split_type` is in `SHUFFLE_SPLIT_TYPES = ["uniform", "stratified"]`. Then if `AutoML._state.retrain_final` is true then we call [AutoMLState._train_with_config](https://github.com/microsoft/FLAML/blob/cdc4201188565e74507c026f86abe540ade4cb72/flaml/automl/automl.py#L3766) which leads to `AutoMLState._prepare_sample_train_data` which leads to the exception.

[to the exception here](https://github.com/microsoft/FLAML/blob/main/flaml/automl/automl.py#L340) because `sample_weight_all` is not set.

### LINKED ISSUE #1553: [Bug]: holdout + sample_weight + time/group split raises ValueError due to missing elif in prepare_data branching
### Describe the bug

When using AutoML.fit(...) with eval_method="holdout" + split_type="time" (or "group") + a sample_weight argument, prepare_data splits the dataset twice and raises:

ValueError: Found input variables with inconsistent numbers of samples: [200, 200, 180]
The first split happens in the if split_type == "time": (or group) branch in flaml/automl/task/generic_task.py (around line 913 / 956), which correctly subsets state.fit_kwargs["sample_weight"] along with X_train and y_train. The code then also enters the elif self.is_classification(): (line 967) or elif self.is_regression(): (line 1011) branch, which calls _train_test_split and tries to split the already-subset sample_weight against the still-full X_train_all / y_train_all, producing the size mismatch.

### Steps to reproduce

```python
import numpy as np
from sklearn.datasets import make_regression
from flaml import AutoML

X, y = make_regression(n_samples=200, n_features=10, n_informative=5, random_state=42)
sample_weight = np.ones(len(y))

automl = AutoML()
automl.fit(
    X_train=X, y_train=y,
    sample_weight=sample_weight,
    split_type="time",
    eval_method="holdout",
    task="regression",
    time_budget=-1,
    max_iter=2,
    estimator_list=["lgbm"],
)
```
Same error reproduces with task="classification" (using make_classification) and with split_type="group" (with groups=...).

### Model Used

_No response_

### Expected Behavior

**Expected behavior**
AutoML.fit(...) completes successfully — prepare_data should split the data exactly once.

**Actual behavior**

ValueError: Found input variables with inconsistent numbers of samples: [200, 200, 180]
  at sklearn/utils/validation.py:464
  via flaml/automl/task/generic_task.py:324  (_train_test_split)
  via flaml/automl/task/generic_task.py:1012 (the elif self.is_regression() branch in prepare_data)

**Suggested fix**
In flaml/automl/task/generic_task.py, change the standalone if split_type == "group": at line 956 to chain into an elif against the preceding if split_type == "time":. The cleanest shape is to make all the special-split branches mutually exclusive with the classification/regression fallback branches, e.g. by guarding the classification/regression branches with elif split_type not in ("time", "group"): (or by structuring the whole block as a single if/elif chain).

I'm happy to take this on; willing to bundle the fix with a related fix for #887 since both share the same test.

### Screenshots and logs

_No response_

### Additional Information

FLAML version: latest main (verified through c8f4763e, the merge commit of #1552)
Python: 3.12
OS: macOS

immu4989:

Two related bugs in `GenericTask.prepare_data` that surface together whenever a user combines a non-shuffle split (`split_type="time"` or `"group"`) with `sample_weight`:

1. **#1553 (`ValueError: Found input variables with inconsistent numbers of samples`)** — `prepare_data` was splitting the data twice. The `if split_type == "group":` branch at `flaml/automl/task/generic_task.py:956` was a standalone `if`, so after the `time`/`group` branch ran (correctly subsetting `state.fit_kwargs["sample_weight"]`), the chained `elif self.is_classification()` / `elif self.is_regression()` branches also fired and tried to split the already-subset weight against the still-full `X_train_all`. One-character fix: chain the `group` branch off the preceding `time` branch with `elif`.

2. **#887 (`AttributeError: 'AutoMLState' has no attribute 'sample_weight_all'`)** — `AutoMLState.prepare_sample_train_data` reads `self.sample_weight_all` whenever `"sample_weight"` is in `fit_kwargs`, but the attribute was only set inside the `SHUFFLE_SPLIT_TYPES` branch. For `split_type="time"` or `"group"` (or any other non-shuffle path), the attribute was never created, raising the AttributeError during the final retrain step. Fix proposed by @sonichi in #887 (2023): initialize `state.sample_weight_all = sample_weight_full` once before the shuffle branching; the SHUFFLE branch still overwrites it after `shuffle(...)` returns the shuffled weight.

The two bugs surface in the same call (`AutoML.fit(... split_type="time", sample_weight=..., retrain_full=True)`) — the `ValueError` fires first in `prepare_data`, and once that's fixed, the `AttributeError` becomes reachable during the retrain step. So both fixes share a single regression test.

## Related issue number

Closes #887
Closes #1553

## Checks

### DIFF
--- a/flaml/automl/task/generic_task.py
+++ b/flaml/automl/task/generic_task.py
@@ -880,6 +880,7 @@ class GenericTask(Task):
                         y_train_all = np.concatenate([y_train_all, y_train_all[:n][rare_index]])
                     count += rare_count
                 logger.info(f"class {label} augmented from {rare_count} to {count}")
+        state.sample_weight_all = sample_weight_full
         SHUFFLE_SPLIT_TYPES = ["uniform", "stratified"]
         if is_spark_dataframe:
             # no need to shuffle pyspark dataframe
@@ -952,7 +953,7 @@ class GenericTask(Task):
                     ) = self._split_pyspark(state, X_train_all, y_train_all, split_ratio)
                 else:
                     X_train, X_val, y_train, y_val = self._split_pyspark(state, X_train_all, y_train_all, split_ratio)
-            if split_type == "group":
+            elif split_type == "group":
                 gss = GroupShuffleSplit(n_splits=1, test_size=split_ratio, random_state=RANDOM_SEED)
                 for train_idx, val_idx in gss.split(X_train_all, y_train_all, state.groups_all):
                     if data_is_df:
--- a/test/automl/test_split.py
+++ b/test/automl/test_split.py
@@ -52,6 +52,28 @@ def test_time():
     _test(split_type="time")
 
 
+def test_time_split_with_sample_weight():
+    """Regression for #887 and #1553: split_type='time' + sample_weight + holdout."""
+    from sklearn.datasets import make_classification
+
+    X, y = make_classification(n_samples=200, n_features=10, n_informative=5, random_state=42)
+    sample_weight = np.ones(len(y))
+    automl = AutoML()
+    automl.fit(
+        X_train=X,
+        y_train=y,
+        sample_weight=sample_weight,
+        split_type="time",
+        retrain_full=True,
+        eval_method="holdout",
+        task="classification",
+        time_budget=-1,
+        max_iter=2,
+        estimator_list=["lgbm"],
+    )
+    assert automl.model is not None
+
+
 def test_groups_for_classification_task():
     from sklearn.externals._arff import ArffException
 


thinkall (approved):
LGTM. Thank you very much for the PR, @immu4989 .

PR #1557, opened 2026-06-06
fix: FLOW2.reach() handles mixed-type conditional parameters (#903)

### LINKED ISSUE #903: Conditional parameter flow2 crash
Sometimes, when sampling a conditional parameter with different types on the branches, like this:
```python
search_space = {
    'param_s': tune.choice([
        {'param': 'None'},
        {'param': tune.qrandint(10, 100, 10)}
    ]),
}
```
the flow2 searcher crashes within the `reach` method of the `FLOW2` class on [this line](https://github.com/microsoft/FLAML/blob/dbd48c555fcfea4a92753bfc6d5f755222b65107/flaml/tune/searcher/flow2.py#L721) where it does
```python
incumbent1[key] - incumbent2.get(key, np.inf)
```
 with errors like this:
```console
TypeError: unsupported operand type(s) for -: 'float' and 'str'
```

Also, if instead of the `'None'` `str` I use the actual `None` value (of `NoneType`), then the optimizer crashes elsewhere (I can't remember whether in the local or global search it is looking for the value of a parameter that is, actually, never sampled).

immu4989:

`FLOW2.reach` in `flaml/tune/searcher/flow2.py` builds a numeric `delta` array by subtracting two incumbents element-wise:

```python
delta = np.array([incumbent1[key] - incumbent2.get(key, np.inf) for key in self._tunable_keys])
```

When a search space contains a conditional `tune.choice` whose branches yield values of different types — e.g. one branch returns the literal string `'None'` and another returns an int from `tune.qrandint` (the exact pattern from #903) — two FLOW2 instances can hold incumbents whose values are different types for the same key. The subtraction then raises:

```
TypeError: unsupported operand type(s) for -: 'float' and 'str'
```

This was reported by @bbudescu in #903 (2023-01); @sonichi's suggested fix in that thread is to wrap the line in `try/except TypeError` and return `False`. This PR implements exactly that. Returning `False` is the right semantics here — a string-vs-numeric incumbent comparison is structurally not "within step" of each other, so reporting "cannot reach" is correct.

## Verified

- Repro from #903 reproduces cleanly on current main via `FLOW2.reach()` (`TypeError: unsupported operand type(s) for -: 'float' and 'str'`).
- After the fix, `reach()` returns `False` for the mixed-type case (no crash).
- A focused regression test was added in `test/tune/test_searcher.py::test_flow2_reach_mixed_type_incumbents` — verified to fail on main without the fix and pass with it.
- Existing tests in `test/tune/test_searcher.py` continue to pass.
- `pre-commit run --files flaml/tune/searcher/flow2.py test/tune/test_searcher.py` — all hooks pass.

## Related issue number

Closes #903

## Checks

### DIFF
--- a/flaml/tune/searcher/flow2.py
+++ b/flaml/tune/searcher/flow2.py
@@ -674,5 +674,8 @@ class FLOW2(Searcher):
             # unordered cat choice is hard to reach by chance
             if config1[key] != config2.get(key):
                 return False
-        delta = np.array([incumbent1[key] - incumbent2.get(key, np.inf) for key in self._tunable_keys])
+        try:
+            delta = np.array([incumbent1[key] - incumbent2.get(key, np.inf) for key in self._tunable_keys])
+        except TypeError:
+            return False
         return np.linalg.norm(delta) <= self.step
--- a/test/tune/test_searcher.py
+++ b/test/tune/test_searcher.py
@@ -345,5 +345,25 @@ def test_unresolved_search_space(caplog):
     ), "BlendSearch should not produce warning about unresolved search space"
 
 
+def test_flow2_reach_mixed_type_incumbents():
+    """Regression test for #903: FLOW2.reach() must return False instead of raising
+    TypeError when conditional choice parameters lead to mixed-type incumbents."""
+    from flaml.tune.searcher.flow2 import FLOW2
+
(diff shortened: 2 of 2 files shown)


thinkall (approved):
Thank you very much for the PR, @immu4989 !

whiskyboy (approved):

PR #1584, opened 2026-08-18
fix(examples): compute the discrete-label threshold from the training split only

### LINKED ISSUE #1175: leaking data  in 4. Forecasting Discrete Values in automl_time_series_forecast
leaking data  in 4. Forecasting Discrete Values in automl_time_series_forecast
https://github.com/microsoft/FLAML/blob/main/notebook/automl_time_series_forecast.ipynb
df["above_mean_sales"] = np.where(df["Sales"] > df["Sales"].mean(), 1, 0)

you put to train  data information from test  data

AmirF194:

In the "Forecasting Discrete Values" example, `above_mean_sales` is thresholded on `df["Sales"].mean()` computed over the full 180-day series, before the train/test split happens a few lines later. That mean is derived in part from the 30-day period held out as `discrete_test_df`, so the training labels carry information from the data the model is later evaluated against.

This PR splits into `discrete_train_df` / `discrete_test_df` first, computes the threshold from `discrete_train_df["Sales"]` only, and applies that single value to both frames. Same fix applied to both the notebook and its mirrored markdown doc under `website/docs/Examples/`.

**Verification:**
- Reproduced the leak with the notebook's own data generator (`hcrystalball.utils.get_sales_data`, same call the cell uses): the full-series mean (17020.14) differs from the train-only mean (16833.40) and changes at least one training row's label.
- Ran the fixed cell logic standalone (same library calls, outside the notebook) and confirmed `discrete_X_train`/`discrete_X_test` keep their original shapes (150/30 rows) and `discrete_train_df` keeps its original column set.
- Did not re-run the notebook's downstream `AutoML.fit`/`predict` cells (no stored outputs exist for the cell I changed, and no CI test executes this notebook), so I can't speak to any effect on the reported accuracy number further down, only that the label computation itself no longer touches the test period.
- Checked the docs mirror (`website/docs/Examples/AutoML-Time series forecast.md`) for the same pattern and fixed it identically; ran `pre-commit` on both changed files (`mdformat` reflowed two lines in the doc, otherwise clean).

## Related issue number

Fixes #1175

## Checks

- [x] I've used pre-commit to lint the changes in this PR.
- [x] I've included the doc change needed (the markdown mirror of this notebook cell had the same leak).
- [ ] I've added tests (no test in this repo executes `automl_time_series_forecast.ipynb`; the fix was verified by running the cell's logic standalone, see above).


AmirF194:

AmirF194:
No rush, just checking in since it's been about a week. Happy to rebase or adjust scope if that would help get it reviewed.

thinkall (changes requested):
Two issues remain before merging:

1. The corrected threshold changes the training labels, but the notebook still contains downstream outputs generated with the old labels, including the model/predictions and `accuracy = 1.0`. Please rerun cells 23–30 or clear those outputs.
2. The comment “No leak into `discrete_test_df`” describes the leakage direction incorrectly. The fix prevents test-period sales from influencing the training labels; the training-derived threshold is intentionally applied to the test frame. Please correct this wording in both the notebook and markdown mirror.

AmirF194:
Both addressed. Cleared the outputs on the fit/predict/accuracy cells rather than rerunning them, since I didn't have hcrystalball installed in this environment and didn't want to commit numbers from a run you couldn't check against.

Reworded the comment: the threshold is what stays training-only, the test frame reusing that same mean is intentional, not something the fix guards against.

thinkall (approved):
The new commit resolves the stale notebook outputs and corrects the leakage-direction wording. I found no remaining issues in the updated changes.

PR #1586, opened 2026-08-24
Fix misaligned sample_weight in group holdout split

shoemoney:

With `split_type="group"`, `eval_method="holdout"`, and `sample_weight` set, the group branch of `_prepare_data` (flaml/automl/task/generic_task.py:956-966 on main) splits X, y, and groups by `train_idx`/`val_idx` but leaves `state.fit_kwargs["sample_weight"]` at full length and never sets `state.weight_val`. `AutoMLState.prepare_sample_train_data` (flaml/automl/state.py:241) then slices the weights positionally, `weight[:sample_size]`, so every trial trains with weights that belong to other rows: rows after the first held-out group get a neighbor's weight, and the held-out rows' weights land on training rows. Because `weight_val` stays `None`, the validation loss that drives model selection ignores sample weights entirely. All of this is silent, no error is raised, and the wrong model can be selected.

#1554 fixed the crash in this configuration (#887, #1553) by chaining the `"group"` branch as an `elif` and setting `state.sample_weight_all`. It did not change how the weights are split. The pattern this PR follows is the pre-existing one: the `"time"` branch (generic_task.py:913-955) and `_train_test_split` (generic_task.py:309) both split weights into `state.fit_kwargs["sample_weight"]` and `state.weight_val`. The `"group"` branch is the only holdout branch that does not. This PR mirrors that pattern inside the `GroupShuffleSplit` loop, handling both `pd.Series` and array weights.

The regression test encodes each row's expected weight in the row itself (column 0 is a row id, weight is 1000 + id) and spies on `RandomForestClassifier.fit`, so any misalignment between the rows an estimator receives and the weights it receives fails the assert; it also asserts `weight_val` is populated. Without the source change the test fails on the alignment assert; with it, `test/automl/test_split.py` passes 9/9.

Not covered here: the group holdout path still does not handle Spark dataframes (the time branch has an explicit `_split_pyspark` arm, the group branch never did, and `GroupShuffleSplit` cannot split a psDataFrame). That is a pre-existing limitation, and this PR keeps the same pandas/numpy scope as the bug.

## Related issue number

Follow-up to #1554 (issues #887, #1553), which stopped the crash in this configuration but left the group branch's weights unsplit.

## Checks

### DIFF
--- a/flaml/automl/task/generic_task.py
+++ b/flaml/automl/task/generic_task.py
@@ -964,6 +964,15 @@ class GenericTask(Task):
                     y_train, y_val = y_train_all[train_idx], y_train_all[val_idx]
                     state.groups = state.groups_all[train_idx]
                     state.groups_val = state.groups_all[val_idx]
+                    if "sample_weight" in state.fit_kwargs:
+                        # NOTE: _prepare_data is before kwargs is updated to fit_kwargs_by_estimator
+                        weight = state.fit_kwargs["sample_weight"]
+                        if isinstance(weight, pd.Series):
+                            state.fit_kwargs["sample_weight"] = weight.iloc[train_idx]
+                            state.weight_val = weight.iloc[val_idx]
+                        else:
+                            state.fit_kwargs["sample_weight"] = weight[train_idx]
+                            state.weight_val = weight[val_idx]
             elif self.is_classification():
                 # for classification, make sure the labels are complete in both
                 # training and validation data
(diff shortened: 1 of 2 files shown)


shoemoney:
Done in ddd01f1: the else-branch now converts weight to ndarray before indexing (so a list/tuple no longer raises TypeError) and skips splitting when weight is None; added a second regression test using a pd.Series with a non-default index to confirm the split stays positional.

thinkall (approved):
The sample-weight split now handles ndarray/list-like values and non-default-index Series positionally, and the regression coverage verifies row/weight alignment. I found no remaining PR-specific issues. The current build failures are the shared Statsmodels 0.15 break addressed by #1587.

PR #1599, opened 2026-09-12
Honor unit quantization for float sampling

### LINKED ISSUE #1598: Honor unit quantization for float sampling
Quantized float domains bypass rounding when q=1, returning continuous values instead of the documented integer increments. This affects quniform, qloguniform and qrandn when using the built-in sampler.

```python
from flaml.tune.sample import quniform
print(quniform(2, 10, 1).sample(size=5, random_state=42))
```

betacatsling:

Quantized float domains bypass rounding when q=1, returning continuous values instead of the documented integer increments. This affects quniform, qloguniform and qrandn when using the built-in sampler.

Limit the q=1 fast path to integer domains; float domains use the existing quantization path. Tests cover all three distributions, scalar and batched samples, non-unit steps, and deterministic seeded sampling.

## Validation

6 regressions fail before the fix; 23 tests pass afterward across the new tests, test_sample.py, test_stop.py and test_reproducibility.py. Validation used Python 3.12 on CPU, with Optuna installed and without Ray. Full repository `pre-commit run --all-files` passes (pre-commit 3.7.1, compatible with the local Git version).

## Checks

- [x] Regression tests added.

Fixes #1598.

### DIFF
--- a/flaml/tune/sample.py
+++ b/flaml/tune/sample.py
@@ -456,7 +456,7 @@ class Quantized(Sampler):
         if not isinstance(random_state, _BackwardsCompatibleNumpyRng):
             random_state = _BackwardsCompatibleNumpyRng(random_state)
 
-        if self.q == 1:
+        if self.q == 1 and isinstance(domain, Integer):
             return self.sampler.sample(domain, spec, size, random_state=random_state)
 
         quantized_domain = copy(domain)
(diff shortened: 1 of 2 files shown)


thinkall (approved):
Reviewed the unit-quantized float sampling implementation, boundary behavior, compatibility, and regression coverage. The fix preserves the intended domain semantics and has no blocking issues.

PR #1601, opened 2026-09-12
Compute trial averages from per-metric report counts

### LINKED ISSUE #1600: Compute trial averages from per-metric report counts
Trial averages use the zero-based training_iteration as their denominator. Reporting scores 1, 3, 5 produces an average of 4 rather than 3. Sparse secondary metrics also use the wrong denominator, which can affect get_best_trial(scope="avg").

```python
from flaml import tune
def evaluate(config):
    for value in [1.0, 3.0, 5.0]:
        tune.report(score=value)
a = tune.run(evaluate, config={}, metric="score", mode="min",
             num_samples=1, use_ray=False, verbose=0)
print(a.trials[0].metric_analysis["score"]["avg"])
```

betacatsling:

Trial averages use the zero-based training_iteration as their denominator. Reporting scores 1, 3, 5 produces an average of 4 rather than 3. Sparse secondary metrics also use the wrong denominator, which can affect get_best_trial(scope="avg").

Track the number of reports separately for each metric and use that count in the running average. Tests exercise the public tune.run/report path with both dense and sparse reports.

## Validation

2 regressions fail before the fix; 7 tests pass afterward across the new tests, test_sample.py, test_stop.py and test_reproducibility.py. Validation used Python 3.12 on CPU, with Optuna installed and without Ray. Full repository `pre-commit run --all-files` passes (pre-commit 3.7.1, compatible with the local Git version).

## Checks

- [x] Regression tests added.

Fixes #1600.

### DIFF
--- a/flaml/tune/trial.py
+++ b/flaml/tune/trial.py
@@ -109,13 +109,15 @@ class Trial:
                         "last": value,
                     }
                     self.metric_n_steps[metric] = {}
+                    self.metric_n_reports[metric] = 1
                     for n in self.n_steps:
                         key = f"last-{n:d}-avg"
                         self.metric_analysis[metric][key] = value
                         # Store n as string for correct restore.
                         self.metric_n_steps[metric][str(n)] = deque([value], maxlen=n)
                 else:
-                    step = result["training_iteration"] or 1
+                    self.metric_n_reports[metric] += 1
+                    step = self.metric_n_reports[metric]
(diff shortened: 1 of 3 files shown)


thinkall (approved):
Reviewed the per-metric report-count aggregation change, including sparse and multi-metric reporting behavior and tests. The averaging logic is correctly scoped per metric and has no blocking issues.

PR #1603, opened 2026-09-12
Preserve zero-valued anonymous tuning metrics

### LINKED ISSUE #1602: Preserve zero-valued anonymous tuning metrics
Returning 0 from an objective, or calling tune.report(0), drops the anonymous metric because report tests its truthiness. The default optimizer then raises KeyError: _metric instead of accepting the valid zero loss.

```python
from flaml import tune
tune.run(lambda config: 0.0, config={}, mode="min",
         num_samples=1, use_ray=False, verbose=0)
```

betacatsling:

Returning 0 from an objective, or calling tune.report(0), drops the anonymous metric because report tests its truthiness. The default optimizer then raises KeyError: _metric instead of accepting the valid zero loss.

Use an explicit None check for the optional anonymous metric. Tests exercise returned and directly reported zero values, False, and positive/negative controls through a real tuning run.

## Validation

6 regressions fail before the fix; 15 tests pass afterward across the new tests, test_sample.py, test_stop.py and test_reproducibility.py. Validation used Python 3.12 on CPU, with Optuna installed and without Ray. Full repository `pre-commit run --all-files` passes (pre-commit 3.7.1, compatible with the local Git version).

## Checks

- [x] Regression tests added.

Fixes #1602.

### DIFF
--- a/flaml/tune/tune.py
+++ b/flaml/tune/tune.py
@@ -209,7 +209,7 @@ def report(_metric=None, **kwargs):
             # calling tune.report() outside tune.run()
             return
     result = kwargs
-    if _metric:
+    if _metric is not None:
         result[DEFAULT_METRIC] = _metric
     trial = getattr(_runner, "running_trial", None)
     if not trial:
--- a/test/tune/test_zero_metric.py
+++ b/test/tune/test_zero_metric.py
@@ -1,0 +1,18 @@
+import pytest
+
+from flaml import tune
+from flaml.tune.result import DEFAULT_METRIC
+
+
+@pytest.mark.parametrize("value", [0, 0.0, False, 1.0, -1.0])
+@pytest.mark.parametrize("report_directly", [False, True])
+def test_anonymous_zero_metric_is_recorded(value, report_directly):
+    def evaluate(config):
+        if report_directly:
+            tune.report(value)
+        else:
+            return value
+
+    analysis = tune.run(evaluate, config={}, mode="min", num_samples=1, use_ray=False, verbose=0)
+    assert analysis.trials[0].last_result[DEFAULT_METRIC] == value
+    assert analysis.best_trial is analysis.trials[0]


thinkall (changes requested):
`flaml/tune/tune.py:212` now preserves a zero metric in result storage, but the default FLOW2 optimizer still discards it: `flaml/tune/searcher/flow2.py:425` and `:482` use `if obj:`. After a nonzero incumbent, a valid `0.0` result leaves FLOW2's `best_obj` and `best_config` stale even though analysis selects zero correctly. Change both downstream checks to an explicit missing-value check such as `obj is not None` (while preserving the NaN/inf policy), and add a multi-trial regression that verifies the optimizer incumbent updates to zero.

betacatsling:
Thanks for catching the downstream case. Updated both FLOW2 callbacks (`on_trial_complete` and `on_trial_result`) to use `obj is not None`, so a valid `0.0` result now updates the incumbent. NaN/inf handling is unchanged: `if obj:` never filtered NaN (NaN is truthy), and the NaN/inf selection policy lives in `get_best_trial(filter_nan_and_inf)` at the analysis level.

Added regressions in `test/tune/test_zero_metric.py`:

- `test_flow2_incumbent_tracks_zero_metric`: a two-trial `tune.run` that establishes a nonzero incumbent, then verifies the zero trial becomes both the analysis best trial and FLOW2's incumbent (`best_obj == 0.0`, `best_config` updated).
- `test_flow2_callback_updates_incumbent_to_zero`: per-callback coverage parameterized over `on_trial_complete`/`on_trial_result` and min/max modes, so either guard regressing is caught independently.

All new cases fail on the previous code; the tune test files pass with the change.

thinkall (approved):
Reviewed the fix since my last review. Both FLOW2 callback paths now use explicit `obj is not None` checks, so valid `0.0` objectives update `best_obj`, `best_config`, and the incumbent while preserving the existing missing/invalid-value handling. The regression coverage exercises both callbacks and a multi-trial run. No blocking issues found.

PR #1606, opened 2026-09-13
fix: qrandint/qlograndint never sample the documented inclusive upper bound

### LINKED ISSUE #1605: qrandint/qlograndint's documented inclusive upper bound is never sampled
### Describe the bug

`qrandint`/`qlograndint` document "lower is inclusive, upper is also inclusive (!)", but the upper bound is unreachable in the common case. `Quantized.sample()` has a q=1 fast path that skips quantization and calls the wrapped `Integer` sampler directly, which draws exclusive of `domain.upper` (same contract as plain `randint`). Since `qrandint`'s default is `q=1`, the documented-inclusive upper is silently never sampled.

This isn't just cosmetic: `flaml/automl/time_series/ts_model.py`'s built-in ARIMA and SARIMAX estimators define their p/d/q/P/D/Q search spaces with `tune.qrandint(lower=0, upper=X, q=1)` at 9 call sites, so the shipped time-series AutoML search never explores the top value of any of those hyperparameters.

For q>1 it's also unreliable, just less consistently: the general branch rounds a raw exclusive-upper draw, so whether the nominal upper bound is reached depends on the rounding fraction.

### Steps to reproduce

```python
from flaml.tune.sample import qrandint
import numpy as np

rs = np.random.RandomState(0)
vals = {qrandint(1, 10).sample(spec=None, random_state=rs) for _ in range(20000)}
print(max(vals))  # 9, never 10

rs = np.random.RandomState(0)
vals = {qrandint(0, 10, 2).sample(spec=None, random_state=rs) for _ in range(20000)}
print(max(vals))  # 8, never 10
```

### Expected Behavior

`qrandint(1, 10)` should sample 10 at least occasionally, matching its own docstring.

### Additional Information

FLAML version: current main (f9e087c). I have a fix ready (extends the sampled range by one quantization step for `Integer` domains, then clamps) and I'll open a PR against this issue.

AmirF194:
Fixes #1605

`Quantized.sample()`'s q=1 fast path bypasses quantization and calls the wrapped `Integer` sampler directly, which draws exclusive of `domain.upper` (matching plain `randint`'s own exclusive-upper contract). Since `qrandint`'s default is q=1, its documented inclusive upper bound was unreachable. For q>1 the general branch rounds a raw exclusive-upper draw, so whether the nominal upper is reached depends on the rounding fraction, e.g. `qrandint(0, 10, 2)` topped out at 8.

Fixed by, for `Integer` domains only, extending the sampled range by one quantization step before drawing and clamping the result back to the true upper bound afterwards. Plain `randint`/`lograndint` never go through `Quantized` at all, so they're untouched, and I left the Float q=1 fast path alone since #1599 is already narrowing that one.

One thing to flag: this touches the same `if self.q == 1:` line as #1599. Whichever lands first, the other will need a small rebase.

Added tests for qrandint at q=1 (default and explicit) and q=2, and qlograndint at q=1: none of them can reach the upper bound on main, all do here. Also a control test that plain randint keeps its exclusive upper. test/tune/test_constraints.py exercises qrandint inside a real BlendSearch run and still passes.


thinkall (changes requested):
Two correctness issues remain:

1. `flaml/tune/sample.py:459`: routing the default `q=1` path through float quantization changes batched samples from integer NumPy arrays to lists of `np.float64`, and can lose precision above `2**53`. Preserve an integer-only `q == 1` path and add scalar/batched type and large-bound tests.
2. `flaml/tune/sample.py:470-474`: extending the sampler by `q` draws values outside the requested domain and clamps them into the upper endpoint, biasing its probability (also in log space). Extend an integer sampler's exclusive upper bound by one, or sample quantization-grid indices directly, and add deterministic edge-bin/distribution tests.

AmirF194:
Both real. Pushed fd63fcdd: instead of rounding a raw draw and clamping, it now draws the quantization grid point's index directly, so the q=1 default stays a plain int (no float division) and every grid point including the top one gets exactly one raw index instead of the top bin absorbing the clamped overshoot.

Verified the bias with a fixed seed: the old code put 40164 of 200000 draws in the top bin of a 5-bin grid against ~39800 for the others, expected 40000 each. New code comes back within 0.5% on every bin. Added that as a deterministic test, plus scalar/batched dtype checks and a large-bound (2**60) precision test. Ran the full test/tune/test_sample.py suite locally, 9 passed.

thinkall (changes requested):
The new commit fixes the previous out-of-domain endpoint clamping, but three blockers remain:

1. `flaml/tune/sample.py:469-478`: computing grid bounds with `domain.lower / self.q` converts integers to float. For nonzero bounds above `2**53`, scalar sampling can return a value below the requested domain and batched sampling can overflow on Windows. Calculate ceiling/floor grid indices using integer division and keep offset arithmetic in Python/safe integer types; add scalar and batched large nonzero-bound tests.
2. `flaml/tune/sample.py:478`: the `q=1` batched compatibility issue remains. Historically `size > 1` returned an integer NumPy array; the new path returns a Python list. Preserve the ndarray shape/dtype and assert the container and dtype in the regression test.
3. `flaml/tune/sample.py:472-478`: rebasing the grid to indices starting at `1` changes `qlograndint` probabilities whenever `lower > q`. For example, `qlograndint(10, 20, 1)` samples log-uniformly over `1..11` and shifts by nine instead of sampling over the actual `10..20` values. Sample absolute grid indices (or weight the actual grid values) and add a deterministic distribution test with `lower > q`.

AmirF194:
Pushed b7d9bfb.

1/2: replaced the float division with integer ceiling/floor division for the grid bounds, and the batched path now returns an int64 ndarray (matches test_qrandint_returns_plain_int_scalar_and_batched and a new test_qrandint_large_nonzero_lower_bound_batched).

3: sampled the grid's own index (value // q) instead of one rebased to start at 1. LogUniform is scale-invariant under multiplication but not under an additive shift, so the rebase was the actual bug. Added test_qlograndint_stays_log_uniform_when_lower_much_greater_than_q, which compares the empirical mean of qlograndint(1000, 1010, 1) against the analytic LogUniform(1000, 1010) mean. Checked it discriminates: reverting just the sample.py change fails that test with a 2.0 gap against a 0.1 tolerance (the rebased implementation's mean sits about 200 standard errors off).

Full suite (11 tests) passes in the repo's own Docker command.

thinkall (changes requested):
The latest commit resolves the large-integer precision issue, restores the `q=1` batched ndarray contract, and samples `qlograndint` from absolute grid values. One compatibility blocker remains:

- `flaml/tune/sample.py:484`: the unconditional ndarray return also changes batched `qrandint`/`qlograndint` with `q > 1` from their historical Python list return to `np.ndarray`. Preserve the exact integer-grid calculation, but return an ndarray only for `q == 1` and retain the list container for `q > 1`. Update `test/tune/test_sample.py:100-102` to assert the two distinct historical contracts rather than codifying the broader return-type change.

AmirF194:
Rebased onto main, it conflicted with #1599 on the same line. Both PRs were carving out opposite exceptions to the same q==1 fast path (yours excluded Float, mine excluded Integer), so the correct merge removes the fast path for both types and always routes through quantization. Pushed 43f17992; all 29 tests pass locally including your new test_unit_quantization.py.

thinkall (changes requested):
The main-branch integrations preserve the inclusive-bound, integer-precision, and log-space fixes, but the remaining compatibility blocker is unchanged:

- `flaml/tune/sample.py:481`: batched integer quantization unconditionally returns `np.ndarray`. Historically `q == 1` returned an integer ndarray, while `q > 1` returned a Python list through `list(quantized)`. Preserve the ndarray only for `q == 1`, retain the list contract for `q > 1`, and update `test/tune/test_sample.py:101` so it does not codify the regression.

AmirF194:
Right, that unification was the actual gap. Pushed 8c4a8c0: q == 1 still returns the ndarray, q > 1 now returns a plain list via list(values), and the two tests that had assumed ndarray for a q > 1 domain now check the list container instead.

thinkall (changes requested):
The `q > 1` list contract is restored, but one `q == 1` compatibility issue remains:

- `flaml/tune/sample.py:476-486`: converting `indices` to `np.int64` before the `q == 1` branch changes the wrapped sampler's historical ndarray dtype. On Windows, the prior `qrandint(..., size>1)` path returned `int32`; the current head forces `int64`. Return `indices` directly for batched `q == 1`, and perform the `int64` conversion/multiplication only for the `q > 1` list path. Add a regression assertion matching the underlying sampler's dtype.

AmirF194:
Right, forcing int64 there was unnecessary. Pushed 2ecd58a: q == 1 now returns `indices` unchanged (no dtype conversion), and the int64 cast plus multiply only happens on the q > 1 list path. Added a test that swaps in a fake int32 sampler and asserts the returned array keeps that dtype; it fails against the previous head.

thinkall (changes requested):
The previous `q == 1` dtype blocker is resolved: returning `indices` preserves the wrapped sampler's platform-specific integer ndarray dtype. One large-domain blocker remains:

- `flaml/tune/sample.py:487`: batched `q > 1` converts indices to `np.int64` before multiplication. For grids above `2**63 - 1`, multiplication wraps and can return zero or negative values outside the requested domain, while scalar sampling remains correct. Construct the historical list by multiplying each index as a Python `int`, and extend the large batched `qrandint`/`qlograndint` tests beyond signed 64-bit range.

AmirF194:
Batched q > 1 now multiplies each grid index as a Python int instead of casting through np.int64, so it no longer wraps above 2**63 - 1. Added test_qrandint_batched_beyond_int64_range (q=2**62, upper=2**65); against the prior code it fails with a negative wrapped value, e.g. -4611686018427387904. Full 13-test suite plus ruff check pass in the repo's own Docker verify command. Pushed 6142c343.

thinkall (approved):
Reviewed the changes since my last review. Batched `q > 1` multiplication now stays in Python integers, preventing signed-64-bit overflow while preserving the historical contracts: scalar Python integers, `q == 1` ndarray with the underlying platform dtype, and `q > 1` Python lists. Inclusive bounds, negative/non-divisible domains, large-integer precision, and actual-grid `qlograndint` behavior are covered and pass. No blocking issues found.

flaml/tune/sample.py
```diff
@@ -1,613 +1,643 @@
-# Copyright 2020 The Ray Authors.
```
jianglibigdata, line 1: Why all the original code is removed? There seems to be some formatting issue. Could you please check and fix the diff? Thanks.

jianglibigdata (changes requested):
The diff doesn't look right. All the existing code is removed and the actual diff is not clear.

AmirF194:
Good catch, that was a real problem. A merge from main earlier in this branch's history flattened this file's line endings from CRLF to LF, so the diff showed the whole file as rewritten. Converted it back to CRLF and pushed; the diff is now the actual ~30-line change plus the new test file.

thinkall (approved):
Reviewed the updates since my prior approval. The author commit changes only line endings, and the merge from `main` does not touch the sampling implementation or tests. The approved scalar/container/dtype contracts, large-integer Python arithmetic, inclusive bounds, precision, and log-grid behavior remain intact. No blocking issues found.

flaml/tune/sample.py
thinkall: Verified at head 46cf97c: the source file again uses CRLF line endings, matching the base, and the diff now isolates the intended implementation changes. The whole-file formatting issue is addressed; resolving this thread.

test/tune/test_sample.py
```diff
@@ -30,3 +34,178 @@ def test_sampler():
...
+    # {1..11}, which for this domain has an expected value around 1003.4 instead of the
+    # analytic LogUniform(1000, 1010) mean below; the gap (~1.6) is over 200 standard
+    # errors of the mean at this sample size, so this is not a flaky comparison.
+    lower, upper, q = 1000, 1010, 1
+    rs = np.random.RandomState(0)
+    n = 200000
+    samples = [qlograndint(lower, upper, q).sample(spec=None, random_state=rs) for _ in range(n)]
+    empirical_mean = sum(samples) / n
+    analytic_mean = (upper - lower) / np.log(upper / lower)
+    assert abs(empirical_mean - analytic_mean) < 0.1, (empirical_mean, analytic_mean)
```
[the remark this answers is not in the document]
thinkall: The discrete probabilities stated here are correct, but the claimed half-unit discrepancy is not. Summing k * log((k+1)/k) / log(1011/1000) for k=1000..1010 gives 1004.9900546207962; the test uses 10/log(1010/1000) = 1004.9917080713044. The difference is only 0.00165345, within the 0.1 tolerance. The seed-0 sample mean is 1005.000235 and the test passes. Extending the upper draw boundary to 1011 approximately offsets flooring. An exact discrete expectation would be a clarity improvement, but this failure claim is invalid; resolving.

flaml/tune/sample.py
```diff
@@ -456,8 +456,38 @@ def sample(
...
+            # rebased to start at 1: LogUniform is scale-invariant under multiplication
+            # (X ~ LogUniform(a, b) implies qX ~ LogUniform(qa, qb)) but not under an
+            # arbitrary additive shift, so rebasing to 1 turned qlograndint's draw into a
+            # shifted, differently-shaped distribution whenever lower > q. Every point, top
+            # one included, is still reachable through exactly one index. Integer
+            # arithmetic throughout (no `/`) avoids the float-division precision loss that
+            # bit scalar and batched sampling above 2**53.
+            q = int(self.q)
+            lower_idx = -(-int(domain.lower) // q)  # ceiling division
+            upper_idx = int(domain.upper) // q  # floor division
```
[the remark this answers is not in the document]
thinkall: The quantization contract requires multiples of q within the bounds. For qrandint(0,10,3), the valid grid is {0,3,6,9}; returning 10 via clamping would introduce an off-grid value. Inclusive bounds make the endpoint eligible when it lies on the grid, rather than requiring an off-grid endpoint. The base implementation also aligned the upper bound down to the grid. The PR description could clarify its outdated clamp wording, but the proposed ceiling-and-clamp change would break quantization semantics. Resolving this correctness claim as invalid.

flaml/tune/sample.py
```diff
@@ -456,8 +456,38 @@ def sample(
         if not isinstance(random_state, _BackwardsCompatibleNumpyRng):
             random_state = _BackwardsCompatibleNumpyRng(random_state)
 
-        if self.q == 1 and isinstance(domain, Integer):
-            return self.sampler.sample(domain, spec, size, random_state=random_state)
+        if isinstance(domain, Integer):
```
thinkall: Concrete fix for the custom-sampler contract issue in https://github.com/microsoft/FLAML/pull/1606#discussion_r4038696187: restrict the grid-index transformation to the two built-in integer samplers. Exact type matching is intentional: a custom subclass may override `sample()` and must not be assumed to return grid indices.

Preserve the historical `q == 1` passthrough for other samplers; for `q > 1`, custom samplers fall through to the existing actual-value quantization below this branch.

```suggestion
        uses_integer_grid = isinstance(domain, Integer) and type(self.sampler) in (
            Integer._Uniform,
            Integer._LogUniform,
        )
        if self.q == 1 and not uses_integer_grid:
            return self.sampler.sample(domain, spec, size, random_state=random_state)

        if uses_integer_grid:
```

Validated this replacement against the current head in memory: a custom sampler returning 12 on Integer(0, 20) retains 12 (rather than 24) with q=2; q=1 preserves its ndarray batch output, q=2 preserves the historical list output, and built-in inclusive uniform grids still work. Please add regression coverage for a custom sampler's observed bounds and returned values for both scalar and batched calls, including a custom subclass of Integer._Uniform.

AmirF194:
Applied the suggested fix as given: `uses_integer_grid` gated on exact
sampler type, q == 1 calls the sampler directly on the real domain,
q > 1 falls through to the existing float quantization. Pushed
6320d970.

Added the two regression tests: a custom subclass of `Integer._Uniform`
(exact-type check treats it as non-built-in on purpose) that records
the domain it's called with, one for q == 1 and one for q == 2. Both
fail against the pre-fix code with the sampler seeing the grid-index
domain (0, 21) instead of the real (0, 20). Full suite is 15 passed,
pre-commit clean.

AmirF194:
CI caught a regression from the last push: the q == 1 passthrough I widened to gate on `uses_integer_grid` also matched Float domains, so `quniform`/`qloguniform`/`qrandn` at q == 1 stopped rounding to the nearest unit (test_unit_quantization.py, 6 of its 18 cases). Scoped the passthrough back to Integer domains only, which is what it gated before this round.

Verified: reverted only sample.py and reran the failing parametrizations, all 6 fail with the exact unrounded-vs-rounded mismatch; with the fix, the full test_sample.py + test_unit_quantization.py suite (33 tests) passes in a clean container. Pushed c095728e.

jianglibigdata (approved):

AmirF194:
Thanks for the thorough review rounds on this one, @thinkall. The q==1/q>1 contract split was the part I kept getting wrong, and your catches on the int64 overflow and CRLF diff noise both landed real bugs.

